PR #3, opened 2021-05-28
ci: Update circleci config for monorepo builds

jacsamell:


.circleci/circle_trigger.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
```
jacsamell, line 1: Note this is sourced from https://github.com/labs42io/circleci-monorepo so doesn't need an in depth review

.circleci/config.yml
```diff
@@ -0,0 +1,786 @@
+version: 2.1
```
jacsamell, line 1: This is the merged version of the other 3 individual repo builds for Circle

matt-twomey (approved):

PR #8, opened 2021-06-04
fix: max group size not stored

matt-twomey:
- Now storing group size and updating group size attribute when retrieving stored groups
- Added a make command to install all node dependencies
- Moved the PR template to the root of the repo (I think this will kick in for the next PR?)

**To Test**
- Create a new group, reuse that group and then change the details of the group using the confirmation page change links - this should now work as expected.


repos/fdbt-site/src/data/auroradb.ts
```diff
@@ -621,8 +621,9 @@ export const insertPassengerType = async (
     passengerType: PassengerType | CompanionInfo[],
     name: string,
     isGroup: boolean,
+    maxGroupSize?: string,
 ): Promise<void> => {
-    const contents = JSON.stringify(passengerType);
+    const contents = maxGroupSize ? JSON.stringify({ maxGroupSize, passengerType }) : JSON.stringify(passengerType);
```
jacsamell, line 626: Would it be better to drive this from isGroup? 
Wonder if actually we should split the group stuff out now as you'd always want to provide maxGroupSize is isGroup is true?

jacsamell (approved):

PR #62, opened 2021-07-13
feat: CFD-295 add the global settings page

JwanKhalaf:
# Description

This ticket creates the user interface for the global settings page and also displays a link to "operator settings" on the home page provided it's not hosted on the production environment.

# Testing instructions

Spin the site up and head to http://localhost:5555/globalSettings and check it looks like the ticket says it should

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix

### DIFF
--- a/repos/fdbt-site/src/pages/globalSettings.tsx
+++ b/repos/fdbt-site/src/pages/globalSettings.tsx
@@ -22,7 +22,7 @@ const GlobalSettings = (): ReactElement => (
                                     <li className="app-subnav__section-item app-subnav__section-item--current">
                                         <a
                                             className="app-subnav__link govuk-link govuk-link--no-visited-state govuk-link--no-underline"
-                                            href="/components/accordion/"
+                                            href="/globalSettings"
                                         >
                                             Settings overview
                                         </a>
@@ -31,7 +31,7 @@ const GlobalSettings = (): ReactElement => (
                                     <li className="app-subnav__section-item">
                                         <a
                                             className="app-subnav__link govuk-link govuk-link--no-visited-state govuk-link--no-underline"
-                                            href="/components/back-link/"
+                                            href="/globalSettings"
                                         >
                                             Passenger types
                                         </a>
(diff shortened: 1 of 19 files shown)


repos/fdbt-site/src/pages/globalSettings.tsx
```diff
@@ -0,0 +1,106 @@
...
+                        <p className="govuk-body">This is a paragraph inside a two-thirds wide column</p>
+                    </div>
+                </div>
+            </main>
+        </div>
+    </BaseLayout>
+);
+
+export const getServerSideProps = (ctx: NextPageContextWithSession): { props: {} } => {
+    if (process.env.STAGE === 'prod' && ctx.res) {
```
jacsamell, line 99: Why the check for ctx.res here?

repos/fdbt-site/tests/pages/home.test.tsx
```diff
@@ -5,11 +5,11 @@ import Home from '../../src/pages/home';
 describe('pages', () => {
     describe('home page', () => {
         it('should render correctly', () => {
-            const tree = shallow(<Home multipleOperators />);
+            const tree = shallow(<Home multipleOperators isOnTestEnvironment />);
             expect(tree).toMatchSnapshot();
         });
         it('should render correctly with no multiple operators', () => {
-            const tree = shallow(<Home multipleOperators={false} />);
+            const tree = shallow(<Home multipleOperators={false} isOnTestEnvironment={true} />);
```
jacsamell, line 12: `isOnTestEnvironment` as on line 8 is the same as `isOnTestEnvironment={true}` here. Are you trying to test both scenarios? In which case one of these needs to be `false`

repos/fdbt-site/src/pages/globalSettings.tsx
RobCatton: Because it can be undefined - we do it on other pages where we do the redirect.

repos/fdbt-site/tests/pages/home.test.tsx
JwanKhalaf: I think we said because this was a temporary thing (until the first bit of global settings makes its way into prod), we wouldn't bother with a unit test. I'm happy to add a test if you like :)

RobCatton (approved):

RobCatton (approved):

PR #71, opened 2021-07-19
feat: cfd-297 check for undefined and array length on map function

JwanKhalaf:
# Description

-   Checking for undefined **and** array length on map function in viewPassengerTypes page.

# Testing instructions

-   Load the `/viewPassengerTypes` page and see if it loads with no errors.

# Type of change

-   [ ] feat - A new feature
-   [x] fix - A bug fix

### DIFF
--- a/repos/fdbt-site/src/pages/viewPassengerTypes.tsx
+++ b/repos/fdbt-site/src/pages/viewPassengerTypes.tsx
@@ -28,13 +28,13 @@ const ViewPassengerTypes = ({ passengerTypes, passengerTypeGroups }: PassengerTy
                             Define age range and required proof documents of your passengers as well as passenger groups
                         </p>
 
-                        {!passengerTypes.length ? (
+                        {!passengerTypes?.length ? (
                             <NoIndividualPassengerTypes />
                         ) : (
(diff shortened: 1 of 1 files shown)


jacsamell (changes requested):
Do we know which of the types is wrong that makes it think none of these should be undefined?

repos/fdbt-site/src/pages/viewPassengerTypes.tsx
```diff
@@ -185,7 +185,7 @@ const PassengerTypeGroups = ({ passengerTypeGroups }: { passengerTypeGroups: Gro
                                     {passengerTypeGroup.maxGroupSize}
                                 </p>
 
-                                {passengerTypeGroup.companions.length
+                                {passengerTypeGroup?.companions?.length
                                     ? passengerTypeGroup.companions.map((companion) => (
```
jacsamell, line 189: Do you need the length check now or can you just do `passengerTypeGroup?.companions?.map`

JwanKhalaf: I think we do. I just did some testing in the Chrome console. An empty array still throws an exception if you try to do `.map()` on it.

PR #73, opened 2021-07-20
Pre Prod Release

matt-twomey:
feat: CFD-297 view passenger types
feat: CFD-267 Point to point period netex
fix: CFD-297 turn on source maps
fix: CFD-297 array check before map call
feat: enable P2P Period and SOP price


JwanKhalaf (approved):
Looks good.

PR #79, opened 2021-07-23
Feature/cfd 357 add name suffix

Aaron-Collis-IW:
# Description

In the produced NeTEx we now include the NaPTAN Indicator for each stop beneath the NeTEX “Name” field as “NameSuffix”

# Testing instructions
These commands should allow you test the code:

- cd ${FDBT_ROOT}/repos/fdbt-netex-output/
- npm run test src/netex-convertor/generateAll.test.ts

- cd ${FDBT_ROOT}/fdbt-dev
- generate-validate-all-parallel-quiet

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix

### DIFF
--- a/repos/fdbt-netex-output/src/netex-convertor/generateAll.test.ts
+++ b/repos/fdbt-netex-output/src/netex-convertor/generateAll.test.ts
@@ -47,6 +47,6 @@ describe('generateAll', () => {
 
         const file1 = await fs.promises.readFile(`../../fdbt-dev/data/generatedNetex/${xmlFileName}`, 'utf-8');
         const file2 = await fs.promises.readFile(`../../fdbt-dev/data/netexData/${xmlFileName}`, 'utf-8');
-        expect(file2).toEqual(file1); //Should this be the other way round?
+        expect(file1).toEqual(file2); 
     });
 });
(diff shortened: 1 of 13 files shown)


repos/fdbt-netex-output/src/netex-convertor/generateAll.test.ts
```diff
@@ -47,6 +47,6 @@ describe('generateAll', () => {
 
         const file1 = await fs.promises.readFile(`../../fdbt-dev/data/generatedNetex/${xmlFileName}`, 'utf-8');
         const file2 = await fs.promises.readFile(`../../fdbt-dev/data/netexData/${xmlFileName}`, 'utf-8');
-        expect(file2).toEqual(file1);
+        expect(file2).toEqual(file1); //Should this be the other way round?
```
RobCatton, line 50: Does the diff look different or less helpful if we swap it round?

repos/fdbt-netex-output/src/netex-convertor/point-to-point-tickets/pointToPointTicketNetexHelpers.ts
```diff
@@ -35,6 +35,7 @@ export const getPointToPointScheduledStopPointsList = (fareZones: FareZone[]): S
         version: 'any',
         id: stop.atcoCode ? `atco:${stop.atcoCode}` : `naptan:${stop.naptanCode}`,
         Name: { $t: stop.stopName },
+        NameSuffix: { $t: stop.indicator },
```
RobCatton, line 38: I think we should be doing something like this 

` NameSuffix: stop.indicator ? { $t: stop.indicator } : null,` 

so that we only create the NameSuffix key if the stop.indicator exists - I think it will create it empty otherwise.

Aaron-Collis-IW: From the data I've looked at, it always exisits however if for whatever reason it doesn't at anypoint this is a good catch. Implemented.

RobCatton (approved):

PR #120, opened 2021-08-19
feat: cfd-313 create time restrictions

matt-twomey:
# Description

-   Allows for the creation of time restrictions in global settings. 

# Testing instructions

-   Add a time restriction and check each of the expected error states.

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix


Aaron-Collis-IW (approved):

repos/fdbt-site/src/design/modules/_layout.scss
```diff
@@ -332,6 +332,7 @@ svg:not(:root) {
 
     .govuk-form-group {
         margin-bottom: 0;
+        max-width: 230px;
```
jacsamell, line 335: This seems like a change that could affect a large number of pages across the site?

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
```diff
@@ -0,0 +1,205 @@
...
+        const startTimeInputs = req.body[`startTime${day}`];
+        const endTimeInputs = req.body[`endTime${day}`];
+        if (isArray(startTimeInputs) && isArray(endTimeInputs)) {
+            const timeBands = startTimeInputs.map((startTime, index) => {
+                return {
+                    startTime: removeAllWhiteSpace(startTime),
+                    endTime: removeAllWhiteSpace(endTimeInputs[index]),
+                };
+            });
+            fullTimeRestrictions.push({
```
jacsamell, line 26: Rather than a forEach with a push to an array, could we change this to a map? I think it would be cleaner and easier to follow?

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
```diff
@@ -0,0 +1,205 @@
...
+
+export const collectInputsFromRequest = (
+    req: NextApiRequestWithSession,
+    timeRestrictionDays: TimeRestrictionDay[],
+): FullTimeRestriction[] => {
+    const fullTimeRestrictions: FullTimeRestriction[] = [];
+    timeRestrictionDays.forEach((day) => {
+        const startTimeInputs = req.body[`startTime${day}`];
+        const endTimeInputs = req.body[`endTime${day}`];
+        if (isArray(startTimeInputs) && isArray(endTimeInputs)) {
```
jacsamell, line 19: Just checking would these always be both an array, or both not an array?

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
```diff
@@ -0,0 +1,205 @@
...
+                isValid24hrTimeFormat(timeBand.endTime) &&
+                timeBand.startTime === timeBand.endTime
+            ) {
+                errors.push(
+                    {
+                        errorMessage: 'Start time and end time cannot be the same.',
+                        id: `start-time-${fullTimeRestriction.day}-${index}`,
+                        userInput: timeBand.startTime,
+                    },
+                    {
```
jacsamell, line 105: Why is there 2 errors here?

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
```diff
@@ -0,0 +1,205 @@
...
+                return unique;
+            }, [] as TimeBand[]),
+        };
+    });
+
+    return cleansedInputs.map((cleansedInput) => {
+        const timeBands: TimeBand[] = [];
+        cleansedInput.timeBands.forEach((timeBand) => {
+            if (timeBand.startTime || timeBand.endTime) {
+                timeBands.push(timeBand);
```
jacsamell, line 143: I think what you want here is a `.filter` instead of forEach

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
```diff
@@ -0,0 +1,205 @@
...
+            timeBands: input.timeBands.reduce((unique, o) => {
+                if (!unique.some((obj) => obj.startTime === o.startTime && obj.endTime === o.endTime)) {
+                    unique.push(o);
+                }
+                return unique;
+            }, [] as TimeBand[]),
+        };
+    });
+
+    return cleansedInputs.map((cleansedInput) => {
```
jacsamell, line 139: This map could be combined with the one above by the looks of it.

jacsamell (commented):
Nice work, few minor comments on code style

repos/fdbt-site/src/design/modules/_layout.scss
matt-twomey: I think because it's a subclass within `.time-restrictions-table` it should only affect form groups that are children of elements with this class? And `.time-restrictions-table` only lives on this page and the original time restrictions page. Does that make sense?

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
matt-twomey: Yep if start time is entered on a row then end time will be entered as `''` even if left blank. I actually think we could force these to always be arrays, even if there's only a single entry, would just mean a bit of reworking - what do you reckon?

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
matt-twomey: Ah yep agreed we could just have one, have changed that.

repos/fdbt-site/src/design/modules/_layout.scss
jacsamell: Ah cool makes sense 👍

repos/fdbt-site/src/pages/api/manageTimeRestriction.ts
jacsamell: Nah that's fine if they are always the same 👍

jacsamell (approved):

PR #192, opened 2021-09-21
feat: service specific products page

RobCatton:
# Description

-   To allow users to see which products they have set against their active BODS services.

# Testing instructions

-   Run locally and then navigate to the /services/pointToPointProducts page, after creating a single/return product against a BODS service, and ensuring the start and end dates line up for that service.

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix


repos/fdbt-site/src/pages/products/pointToPointProducts.tsx
```diff
@@ -0,0 +1,147 @@
...
+    if (!serviceId || isArray(serviceId)) {
+        throw new Error('Unable to find line name to show products for.');
+    }
+    const service = await getBodsServiceByNocAndId(noc, serviceId);
+    const products = await getPointToPointProductsByLineId(noc, service.lineId);
+    const productsToDisplay = filterProductsNotToDisplay(service, products);
+    const formattedProducts = await Promise.all(
+        productsToDisplay.map(async (product) => {
+            const matchingJson = await getMatchingJson(product.matchingJsonLink);
+            const productDescription = `${upperFirst(matchingJson.passengerType)} - ${matchingJson.type} ${
```
jacsamell, line 126: Should matchingJson.type be sentence case?

RobCatton: I don't think so, I think we want it to say "Adult - return" rather than "Adult - Return" as thats more start case. Could be wrong, what do you think?

JwanKhalaf (approved):
Looks good

repos/fdbt-site/src/pages/products/pointToPointProducts.tsx
jacsamell: What about flatFare ? Would have thought this should be 'flat fare' or 'Flat Fare' or something

RobCatton: This page is just point to point products, the ones like flat fare and stuff will use their product name I think, on a different page.

PR #263, opened 2021-11-01
fix: service end date issue

RobCatton:
# Description

-   If a service had no end date, it would use todays date instead. This fix stops that from happening.

# Testing instructions

-   Locally, none of the services have end dates, which makes it easy to test. You can also alter the code if you want to make them have end dates, to see how they look.

# Type of change

-   [ ] feat - A new feature
-   [x] fix - A bug fix


repos/fdbt-site/src/pages/products/services.tsx
```diff
@@ -126,15 +132,27 @@ export const matchProductsToServices = (
...
-        ...service,
-        products:
-            productsByLine
-                .get(service.lineId)
-                ?.filter((product) =>
-                    showProductAgainstService(product.startDate, product.endDate, service.startDate, service.endDate),
-                ).length ?? 0,
-    }));
+    return services.map((service) => {
+        const serviceWithEndDatesSwapped = {
```
matt-twomey, line 136: Could we avoid defining this variable, since it's only used once, and instead override endDate in the return statement?

RobCatton: @matt-twomey i did consider this, but was unsure how to write it. Do you have an example?

matt-twomey (approved):

jacsamell (approved):

PR #338, opened 2021-11-25
feat: temporary tables for reference data import

RobCatton:
# Description

-   Currently we clear out the tables at the beginning of every day - instead of doing this, and having no data if the import fails, we are creating some new temporary tables to hold the data as the import happens.

# Testing instructions

-   Run the new lambda and the reference data lambdas afterwards, to ensure they work in sequence.

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix


repos/exporter/lib/database.ts
```diff
@@ -265,3 +265,88 @@ export const removeAllServicesRequiringAttentionIds = async (): Promise<void> =>
...
+
+    if (percentageResult < 75) {
+        throw new Error(
+            `Reference data import has not completed, as only ${percentageResult}% of yesterday's data has been imported.`,
+        );
+    }
+};
+
+export const deleteAndRenameTables = async (): Promise<void> => {
+    const renameQuery = `BEGIN TRANSACTION
```
jacsamell, line 296: Do you need to disable and reenable the foreign key constraints?

repos/exporter/lib/exporter-stack.ts
```diff
@@ -110,6 +110,39 @@ export class ExporterStack extends cdk.Stack {
...
+                sourceMap: true,
+                sourceMapMode: SourceMapMode.DEFAULT,
+            },
+            timeout: Duration.minutes(15),
+            logRetention: 180,
+        });
+
+        this.addAlarmsToLambda(stage, tableRenameFunction, `tableRenameFunction-${stage}`, 420000);
+
+        tableRenameFunction.addToRolePolicy(new PolicyStatement({ actions: ['ssm:GetParameter'], resources: ['*'] }));
```
joeds13, line 139: Can the resources be restricted here and explicitly list only the parameters it needs, rather than `*`?

repos/exporter/lib/exporter-stack.ts
```diff
@@ -121,43 +152,50 @@ export class ExporterStack extends cdk.Stack {
...
             },
             timeout: Duration.minutes(15),
             logRetention: 180,
-            memorySize: 2048,
         });
 
-        netexBucket.grantReadWrite(zipperFunction);
+        this.addAlarmsToLambda(tableRenameFunction, `tableRenameFunction-${this.stage}`, 420000);
+
+        tableRenameFunction.addToRolePolicy(new PolicyStatement({ actions: ['ssm:GetParameter'], resources: ['*'] }));
```
joeds13, line 188: Can the resources be restricted here and explicitly list only the parameters it needs, rather than `*`?

Same for the other two Lambdas, but appreciate they're not being added by this PR

repos/exporter/lib/exporter-stack.ts
```diff
@@ -1,59 +1,70 @@
...
 
-        const vpc = Vpc.fromLookup(this, 'vpc', { vpcName: `fdbt-vpc-${stage}` });
-        const vpcSubnets = {
-            subnets: [
-                Subnet.fromSubnetId(this, 'vpc-subnet-a', Fn.importValue(`${stage}:PrivateSubnetA`)),
-                Subnet.fromSubnetId(this, 'vpc-subnet-b', Fn.importValue(`${stage}:PrivateSubnetB`)),
-            ],
-        };
+        this.addNeedsAttentionLambda(securityGroup, vpc, vpcSubnets);
+        this.addReferenceDataLambdas(securityGroup, vpc, vpcSubnets);
```
joeds13, line 52: Does it make sense for these to all be in the same CDK Stack?

They seem to be distinct enough parts of the application, otherwise we're heading in the direction of the `ExporterStack` really being the everything stack, we should look to set a good pattern on this from the beginning

joeds13 (changes requested):

repos/exporter/lib/exporter-stack.ts
joeds13: When I paired with Rob we wrote some code to do this, but it seems to have been lost

repos/exporter/lib/exporter-stack.ts
jacsamell: Yeah I've been meaning to rename the stack before I leave now it does more stuff. I'll try and get this done separately

repos/exporter/lib/exporter-stack.ts
jacsamell: Didn't see that code, I'll see if I can find it

repos/fdbt-site/src/pages/salesConfirmation.tsx
```diff
@@ -117,10 +117,12 @@ const SalesConfirmation = ({
...
 
-export const getServerSideProps = (ctx: NextPageContextWithSession): { props: SalesConfirmationProps } => {
+export const getServerSideProps = (
+    ctx: NextPageContextWithSession,
+): GetServerSidePropsResult<SalesConfirmationProps> => {
     const fareTypeAttribute = getSessionAttribute(ctx.req, FARE_TYPE_ATTRIBUTE);
-    if (!fareTypeAttribute && ctx.res) {
-        redirectTo(ctx.res, '/home');
+    if (!fareTypeAttribute) {
+        return { redirect: { destination: '/home', permanent: false } };
```
JwanKhalaf, line 125: How come this change was made?

RobCatton: Basically every time we do a redirect, we also ought to be doing a return (otherwise the code continues and issues occur). In API's its easy, as we just add 'return' on the line after. However, in server side props, theres a particular return type needed to make typescript happy - which is what this is.

JwanKhalaf (approved):

PR #339, opened 2021-11-25
fix/cfd 558 circular return

Aaron-Collis-IW:
# Description

Create a validation error which alerts the user that they can't use a one direction journey for return and point to point period products. 

# Testing instructions

-   On test:
- Use noc FRSV
- Try and create a p2p period and return for the 1a service

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix


repos/fdbt-site/src/pages/api/service.ts
```diff
@@ -17,8 +17,27 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
 
         const service = await getServiceByIdAndDataSource(getAndValidateNoc(req, res), serviceId, dataSource);
+        const directions = Array.from(
+            service.journeyPatterns.reduce((set, pattern) => {
+                set.add(pattern.direction);
+                return set;
+            }, new Set<string>()),
+        );
+
+        if (directions.length === 1 && isReturn) {
```
RobCatton, line 31: What will happen here if we have outbound + clockwise (or a pairing which both mean the same thing)? That would mean there are multiple directions in this array, but they both mean the same thing.

repos/fdbt-site/src/pages/api/service.ts
```diff
@@ -17,8 +17,27 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+            const errors: ErrorInfo[] = [
+                {
+                    id: 'service',
```
RobCatton, line 34: Do we want to pass back the selected service to use as the default selection? The error message mentions a service, but it wont be shown on the screen.

repos/fdbt-site/src/pages/api/service.ts
```diff
@@ -17,8 +17,27 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+                    errorMessage: 'The chosen a service only has one direction and can`t be used for a return',
```
RobCatton, line 35: Error message needs slightly altering, the third word seems unnecessary.

repos/fdbt-site/src/pages/api/service.ts
jacsamell: Yeah good point, we need to check that they have inbound + outbound below 👍

repos/fdbt-site/src/pages/api/service.ts
Aaron-Collis-IW: done

repos/fdbt-site/src/pages/api/service.ts
Aaron-Collis-IW: this is done

repos/fdbt-site/src/pages/api/service.ts
Aaron-Collis-IW: this is done

PR #347, opened 2021-12-01
fix: removing adjacently identical stops from master stop list

RobCatton:
# Description

-   If two bus stops are next to one another and are the same, then we remove the latter one.

# Testing instructions

-  n/a

# Type of change

-   [ ] feat - A new feature
-   [x] fix - A bug fix


repos/fdbt-site/src/utils/apiUtils/matching.ts
```diff
@@ -125,6 +125,22 @@ const validateSequenceNumbers = (stops: StopPoint[]): stops is (StopPoint & { se
...
+
+            if (stop !== stops[index - 1]) {
+                return stop;
+            }
+
+            return '';
+        })
+        .filter((stop) => stop);
+};
+
```
jacsamell, line 143:
```
export const removeDuplicateAdjacentStops = (stops: string[]): string[] => {
    return stops.reduce<string[]>((newStops, stop, index, stops) => {
        if (stop !== stops[index - 1]) {
            newStops.push(stop);
        }
        return newStops;
    }, []);
};
```

Aaron-Collis-IW (approved):

repos/fdbt-site/src/utils/apiUtils/matching.ts
RobCatton: done

PR #349, opened 2021-12-02
feat: cfd-481 - edit fare triangle prices

JwanKhalaf:
# Description

-   Edit fare triangle prices

# Testing instructions

-   For points to points, attempt to edit by uploading a new csv (edit existing csv, for example; change fare stage names and add a new fare stage)

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix

### DIFF
--- a/repos/fdbt-site/src/constants/attributes.ts
+++ b/repos/fdbt-site/src/constants/attributes.ts
@@ -2,10 +2,6 @@
 
 export const FARE_TYPE_ATTRIBUTE = 'fdbt-fare-type';
 
-export const MATCHING_JSON_ATTRIBUTE = 'fdbt-matching-json';
-
-export const PRODUCT_AND_SERVICE_ID_ATTRIBUTE = 'fdbt-product-and-service-id';
-
 export const CARNET_FARE_TYPE_ATTRIBUTE = 'fdbt-carnet-fare-type';
 
 export const PASSENGER_TYPE_ATTRIBUTE = 'fdbt-passenger-type';
@@ -113,3 +109,7 @@ export const GS_REFERER = 'fdbt-global-settings-referer';
 export const GS_OPERATOR_DETAILS_ATTRIBUTE = 'fdbt-global-settings-operator-details';
 
 export const CSV_ZONE_FILE_NAME = 'fdbt-csv-zone-file-name';
+
+export const MATCHING_JSON_ATTRIBUTE = 'fdbt-matching-json';
+
+export const PRODUCT_AND_SERVICE_ID_ATTRIBUTE = 'fdbt-product-and-service-id';
--- a/repos/fdbt-site/src/constants/attributes.ts
+++ b/repos/fdbt-site/src/constants/attributes.ts
@@ -2,6 +2,8 @@
 
 export const FARE_TYPE_ATTRIBUTE = 'fdbt-fare-type';
 
+export const MATCHING_JSON_ATTRIBUTE = 'fdbt-matching-json';
+
 export const CARNET_FARE_TYPE_ATTRIBUTE = 'fdbt-carnet-fare-type';
 
 export const PASSENGER_TYPE_ATTRIBUTE = 'fdbt-passenger-type';
--- a/repos/fdbt-site/src/constants/attributes.ts
+++ b/repos/fdbt-site/src/constants/attributes.ts
@@ -4,6 +4,8 @@ export const FARE_TYPE_ATTRIBUTE = 'fdbt-fare-type';
 
 export const MATCHING_JSON_ATTRIBUTE = 'fdbt-matching-json';
 
+export const PRODUCT_AND_SERVICE_ID_ATTRIBUTE = 'fdbt-product-and-service-id';
+
 export const CARNET_FARE_TYPE_ATTRIBUTE = 'fdbt-carnet-fare-type';
 
 export const PASSENGER_TYPE_ATTRIBUTE = 'fdbt-passenger-type';
--- a/repos/fdbt-site/src/pages/api/updateFareTriangle.ts
+++ b/repos/fdbt-site/src/pages/api/updateFareTriangle.ts
@@ -40,21 +40,17 @@ export const thereIsAFareStageNameMismatch = (fareTriangleData: UserFareStages, 
     return thereIsANameMismatch;
 };
 
-const stageCountMismatchError: ErrorInfo[] = [
-    {
-        id: errorId,
-        errorMessage:
-            'The number of fare stages of your updated fares triangle do not match the one you have previously uploaded. Update your triangle, ensuring the number of fare stages match before trying to upload again',
-    },
-];
+const stageCountMismatchError: ErrorInfo = {
+    id: errorId,
+    errorMessage:
+        'The number of fare stages of your updated fares triangle do not match the one you have previously uploaded. Update your triangle, ensuring the number of fare stages match before trying to upload again',
+};
 
-const nameMismatchError: ErrorInfo[] = [
-    {
-        id: errorId,
-        errorMessage:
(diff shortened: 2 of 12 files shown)


repos/fdbt-site/src/pages/api/updateFareTriangle.ts
```diff
@@ -0,0 +1,342 @@
...
+    res: NextApiResponse,
+    errors: ErrorInfo[],
+    poundsOrPence?: string,
+): void => {
+    updateSessionAttribute(req, CSV_UPLOAD_ATTRIBUTE, { errors, poundsOrPence });
+
+    redirectTo(res, '/csvUpload');
+};
+
+export const containsDuplicateFareStages = (fareStageNames: string[]): boolean =>
```
RobCatton, line 49: if we have some duplicated methods across both files, could we import the duplicated ones from the original file?

JwanKhalaf: Done, good shout.

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
```diff
@@ -0,0 +1,158 @@
...
+
+const errorId = 'csv-upload';
+
+export const config = {
+    api: {
+        bodyParser: false,
+    },
+};
+
+const getNamesOfFareZones = (ticket: WithIds<SingleTicket> | WithIds<ReturnTicket>) => {
```
RobCatton, line 24: Could we make a unit test for this? Where we pass it a Single Ticket and Return Ticket, and just ensure it gets the right results.

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
```diff
@@ -0,0 +1,158 @@
...
+        const combinedNamesOnly = combinedFareZones.map((x) => x.name);
+
+        // get rid of duplicates
+        fareZoneNames = [...new Set(combinedNamesOnly)];
+    }
+
+    return fareZoneNames;
+};
+
+const checkFareStageNames = (fareTriangleData: UserFareStages, fareZoneNames: string[]): boolean => {
```
RobCatton, line 41: Could we make a quick unit test for this? I think only one scenario needed.

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
```diff
@@ -0,0 +1,158 @@
...
+    const thereIsANameMismatch = fareTriangleData.fareStages.some((fs) => {
+        return !fareZoneNames.includes(fs.stageName);
+    });
+
+    return thereIsANameMismatch;
+};
+
+const getStageCountMismatchError = () => {
```
RobCatton, line 49: Does this and the one below need to be a function? Could it instead just be a variable? 
```suggestion
const stageCountMismatchError: ErrorInfo[] = [
        {
            id: errorId,
            errorMessage:
                'The number of fare stages of your updated fares triangle do not match the one you have previously uploaded. Update your triangle, ensuring the number of fare stages match before trying to upload again',
        },
    ];
```

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
```diff
@@ -0,0 +1,158 @@
...
+        },
+    ];
+
+    return errors;
+};
+
+export default async (req: NextApiRequestWithSession, res: NextApiResponse): Promise<void> => {
+    try {
+        const ticket = getSessionAttribute(req, MATCHING_JSON_ATTRIBUTE) as
+            | WithIds<SingleTicket>
```
RobCatton, line 76: Unsure if its a code formatting thing or not, does this type need the first pipe ( | ) after the 'as' ?

JwanKhalaf: It is `es-lint/prettier` that does that.

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
JwanKhalaf: Will add unit test.

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
JwanKhalaf: Will add unit test.

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
```diff
@@ -0,0 +1,158 @@
...
+
+            if (!fareTriangleData || !fareZoneNames) {
+                return;
+            }
+
+            let errors;
+
+            // check to see if the the number of fare stages does not match
+            if (fareTriangleData.fareStages.length !== fareZoneNames.length) {
+                errors = getStageCountMismatchError();
```
RobCatton, line 125: The way this is written, if the errors array gets set here due to there being a different number of fare stages, it will get overwritten on line 131. I think we might need to check to see if errors.length > 0 and if it is, redirect before we do the 2nd error check

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            const thereIsANameMismatch = checkFareStageNames(fareTriangleData, fareZoneNames);
+
+            if (thereIsANameMismatch) {
+                errors = getNameMismatchError();
+            }
+
+            // check to see if we had errors
+            if (errors !== undefined) {
```
RobCatton, line 135: I think we usually do errors.length > 0 in this scenario, but this also works.

repos/fdbt-site/src/constants/attributes.ts
```diff
@@ -2,6 +2,10 @@
 
 export const FARE_TYPE_ATTRIBUTE = 'fdbt-fare-type';
 
+export const MATCHING_JSON_ATTRIBUTE = 'fdbt-matching-json';
```
RobCatton, line 5: Could we move these to the bottom of the file? It makes them a little easier to find if we keep to the order (oldest at top and newest at bottom).

repos/fdbt-site/src/pages/products/productDetails.tsx
```diff
@@ -64,18 +65,49 @@ const ProductDetails = ({
...
+    );
+};
+
 export const getServerSideProps = async (ctx: NextPageContextWithSession): Promise<{ props: ProductDetailsProps }> => {
     const noc = getAndValidateNoc(ctx);
-    const productId = ctx.query?.productId;
+
     const serviceId = ctx.query?.serviceId;
 
+    const productId = ctx.query?.productId;
```
RobCatton, line 97: Could we add the productId check in the if statement below? Would make sense to check them both at the same time, as we should fail immediately if they arent there.

RobCatton: I think also we only need to check if the type isnt a string, because undefined and string[] are both incorrect.

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
JwanKhalaf: Unit test done :+1:

repos/fdbt-site/src/pages/api/updateFareTriangle.ts
JwanKhalaf: Unit test done :+1:

repos/fdbt-site/src/constants/attributes.ts
JwanKhalaf: Done

RobCatton (approved):

PR #350, opened 2021-12-03
feat: cfd-568 - migrate new naptan site

JwanKhalaf:
# Description

-   Handle the new URL and refactor the naptan handler.

# Testing instructions

-   Run the handler via the AWS Console

# Type of change

-   [x] feat - A new feature
-   [ ] fix - A bug fix

### DIFF
--- a/repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
+++ b/repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
@@ -2,31 +2,58 @@ import boto3
 import os
 import logging
 from io import BytesIO
-from zipfile import ZipFile
-from urllib.request import urlopen
+from zipfile import ZipFile, is_zipfile
+import requests
 
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
-s3 = boto3.resource('s3')
+s3 = boto3.resource("s3")
 
 
 def lambda_handler(event, context):
     try:
-        resp = urlopen(os.getenv('DATA_URL'))
-        zipfile = ZipFile(BytesIO(resp.read()))
-        target_file = os.getenv('TARGET_FILE')
-        
-        for filename in zipfile.namelist():
-            if (target_file and filename == target_file) or not target_file:
-                s3.meta.client.upload_fileobj(
-                    zipfile.open(filename),
-                    os.getenv('BUCKET_NAME'),
-                    filename,
-                    ExtraArgs={
-                        'ContentType': os.getenv('CONTENT_TYPE')
-                    }
-                )
+        data_url = os.getenv("DATA_URL")
+        content_type = os.getenv("CONTENT_TYPE")
+        bucket_name = os.getenv("BUCKET_NAME")
+
+        if data_url is None:
+            raise Exception("There was no DATA_URL environment variable!")
+
+        if content_type is None:
+            raise Exception("There was no CONTENT_TYPE environment variable!")
+
+        if bucket_name is None:
+            raise Exception("There was no BUCKET_NAME environment variable!")
+
+        response = requests.get(data_url, allow_redirects=True, timeout=30)
+        file = BytesIO(response.content)
+
+        if is_zipfile(file):
+            logger.info("File retrieved is a ZIP file")
+
+            zip_file = ZipFile(file)
+            target_file = os.getenv("TARGET_FILE")
+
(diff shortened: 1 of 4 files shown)


repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
```diff
@@ -2,31 +2,70 @@
...
+        data_url = os.getenv('DATA_URL')
+        backup_data_url = os.getenv('BACKUP_DATA_URL')
+        content_type = os.getenv('CONTENT_TYPE')
+        bucket_name = os.getenv('BUCKET_NAME')
+
+        if data_url is None:
+            raise Exception("There was no DATA_URL environment variable!")
+
+        if backup_data_url is None:
+            raise Exception("There was no BACKUP_DATA_URL environment variable!")
```
jacsamell, line 24: This will break the NocRetriever

jacsamell (changes requested):

repos/fdbt-reference-data-service/src/retrievers/serverless.yml
```diff
@@ -1,199 +1,200 @@
...
-              treatMissingData: notBreaching
-            - name: functionInvocations
-              treatMissingData: notBreaching
+  NaptanRetriever:
+    handler: csv_retriever/main.lambda_handler
+    timeout: 60
+    memorySize: 512
+    environment:
+      DATA_URL: https://naptan.dft.gov.uk/Download/File/Stops.csv
+      BACKUP_DATA_URL: https://beta-naptan.dft.gov.uk/Download/File/Stops.csv
```
jacsamell, line 119: These should be the other way around I think, the beta one is the one that currently works

repos/fdbt-reference-data-service/src/retrievers/.python-version
```diff
@@ -0,0 +1 @@
+jtest
```
jacsamell, line 1: whats this?

repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
```diff
@@ -2,31 +2,70 @@
...
+                            'ContentType': content_type 
+                        }
+                    )
+        else:
+            logger.info("File retrieved is not a ZIP file")
+
+            target_file = os.getenv('TARGET_FILE')
+
+            if target_file is None:
+                raise Exception("There was no TARGET_FILE environment variable!")
```
matt-twomey, line 65: It looks like target_file may not be set sometimes, as in line 50 above, and we don't raise an exception. Are we sure we want to except if there's no target_file environment variable, in the case that the file isn't a zip file?

repos/fdbt-reference-data-service/src/retrievers/serverless.yml
```diff
@@ -1,199 +1,200 @@
...
+  name: aws
+  region: eu-west-2
+  runtime: python3.8
+  stage: ${opt:stage, 'test'}
+  stackName: ${self:service}-${self:provider.stage}
+  lambdaHashingVersion: 20201221
+  iam:
+    role:
+      statements:
+        - Effect: "Allow"
```
matt-twomey, line 13: Are the formatting changes intentional, or the result of an IDE not using the correct formatter?

repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
Aaron-Collis-IW: MOved this to with in the except

repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
JwanKhalaf: On line 53, it checks for it.

repos/fdbt-reference-data-service/src/retrievers/serverless.yml
Aaron-Collis-IW: The beta one is the one that works but if we swap them around we will use the beta link until it dies, wouldn't it be better to switch to the normal one as soon as it goes live, which is what the current code will do. Does that make sense?

Aaron-Collis-IW (approved):
Looks good, just need to wait for locality name to be added by dft

joeds13 (approved):

PR #419, opened 2022-02-02
fix: cfd-638 retry route selection if route cannot be used for return

Aaron-Collis-IW:
## Description

fix: cfd-638 retry route selection if route cannot be used for return

## Testing instructions

Run ui tests

### DIFF
--- a/repos/fdbt-site/cypress_tests/cypress/support/helpers.ts
+++ b/repos/fdbt-site/cypress_tests/cypress/support/helpers.ts
@@ -528,3 +528,13 @@ export const addFlatFareProductIfNotPresent = (): void => {
         }
     });
 };
+
+export const retryRouteChoiceOnReturnProductError = (): void => {
+    cy.get('main').then(($main) => {
+        if ($main.text().includes('you cannot create a return product for this service')) {
+            cy.log('Retrying, could not create a return product for for the first service');
+            selectRandomOptionFromDropDown('service');
+            continueButtonClick();
+        }
+    });
+};
--- a/repos/fdbt-site/cypress_tests/cypress/support/steps.ts
+++ b/repos/fdbt-site/cypress_tests/cypress/support/steps.ts
@@ -21,6 +21,7 @@ import {
     randomlyChooseProductPeriodValidity,
     clickRandomElementInTable,
     clickElementByText,
+    retryRouteChoiceOnReturnProductError,
 } from './helpers';
 
 export const defineUserTypeAndTimeRestrictions = (): void => {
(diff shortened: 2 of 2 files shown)


repos/fdbt-site/cypress_tests/cypress/support/steps.ts
```diff
@@ -117,6 +118,8 @@ export const completeFlatFareCarnet = (): void => {
 export const completeServicePage = (): void => {
     selectRandomOptionFromDropDown('service');
     continueButtonClick();
+    checkAndRepickService();
+    checkAndRepickService();
```
joeds13, line 122: So it will retry twice, what happens if the first three options all fail?

We probably want to keep iterating over all of the possible options until we run out

repos/fdbt-site/cypress_tests/cypress/support/helpers.ts
```diff
@@ -528,3 +528,13 @@ export const addFlatFareProductIfNotPresent = (): void => {
         }
     });
 };
+
+export const checkAndRepickService = (): void => {
+    cy.get('body').then(($body) => {
+        if ($body.text().includes('There is a problem')) {
```
joeds13, line 534: Is this too vague, will it catch other potential issues we'd immediately want to fail the test on?

Would be good to check against the error detail "As your service only operates in a single direction..."

Aaron-Collis-IW: yeah, I'll make these changes

repos/fdbt-site/cypress_tests/cypress/support/helpers.ts
```diff
@@ -528,3 +528,13 @@ export const addFlatFareProductIfNotPresent = (): void => {
         }
     });
 };
+
+export const checkAndRepickService = (): void => {
+    cy.get('body').then(($body) => {
+        if ($body.text().includes('Tcannot create a return for this service')) {
```
joeds13, line 534: Is `Tcannot` a typo? 

`cannot create a return for this service` does not match the error detail message, or is it trying to match to some other hidden data in the DOM?

The error detail on the page is `As your service only operates in a single direction, you cannot create a return product for this service` we should either do a full match on that to be explicit, or perhaps just the second half `you cannot create a return product for this service` to allow for future other reasons why a return may not be valid for this service.

repos/fdbt-site/cypress_tests/cypress/support/steps.ts
joeds13: We have an opportunity here to change the code to fail 0% of the time, assuming that one of the services setup for the `noc` will allow a return product, so I think that would be more robust that this. 

Unless we just want to ensure this services exists in the test seed data and we can do away with this retry and just choose the one we create from the dropdown.

If we want to continue on the retry approach, relying on Cypress/BrowserStack to retry will not only increase the overall test run time, but unless you can find out if the randomness in a test run is non-deterministic then the same random option might be chosen from the dropdown each time. Therefore we should exhaust all dropdown options in a single test run.

I think a bit of a refactor is needed to make this code flow better and be as robust as possible. We should look to remove the duplication of the lines:
```
    selectRandomOptionFromDropDown('service');
    continueButtonClick();
```
So rather than trying the `continueButtonClick()', then going to the retry functions in the error case, we do something like this pseudo code:
```
completeServicePage(){
  serviceOption = numberOfOptionsInServiceDropdown
  for serviceOption; do
    if "you cannot create a return product for this service" is not on the page;
      selectOptionFromDropDown('service', serviceOption);
      serviceOption - 1
      continueButtonClick();
    fi
   error:
      log("no options left, test failed")
  done
}
```
Where we iterate through all the dropdown options seeing if any will work, logging and erroring if none do

Further reading and other suggestions on dealing with conditional testing can be found here: https://docs.cypress.io/guides/core-concepts/conditional-testing

repos/fdbt-site/cypress_tests/cypress/support/helpers.ts
```diff
@@ -528,3 +528,13 @@ export const addFlatFareProductIfNotPresent = (): void => {
...
+            cy.log('There is a problem, repicking service');
```
joeds13, line 535: The log message should include what the problem is, so something like "Trying to use a Service that cannot have a return product created for, trying another service"

joeds13 (changes requested):

joeds13:
Due to Cypress' implementation and design, we're limited in options for conditional testing. So we've opted to implement a simple retry twice of the random selection, which will nearly always results in a successful step and if not will probably pass when Cypress retries the whole test up to three times too.

joeds13 (approved):

PR #429, opened 2022-02-11
fix: CFD-648 Update descriptions on flat fare ticketRepresentation page

Aaron-Collis-IW:
## Description

Update descriptions on flat fare ticketRepresentation page

## Testing instructions

Test flat fare and period tickets and check on the ticketRepresentation page the correct hints are showed


repos/fdbt-site/src/pages/ticketRepresentation.tsx
```diff
@@ -37,6 +37,26 @@ const getFareTypeDesc = (fareType: TicketType) => {
     }
 };
 
+const getFareTypeHint = (fareType: TicketType) => {
+    switch (fareType) {
+        case 'period':
+            return [
+                'Unlimited travel within a geographical zone',
+                'Unlimited travel on specific service or set of services',
+            ];
```
joeds13, line 46: Rather than relying on list ordering, I think we should return objects here, so we can then lookup in the component by the hint's key, rather than by a positional argument

```javascript
return {
    geoZone: 'Unlimited travel within a geographical zone',
    multipleServices: 'Unlimited travel on specific service or set of services',
};
```
```javascript
{
    value: 'geoZone',
    label: 'A ticket within a geographical zone',
    hint: fareTypeHint.geoZone,
}
```

This will make it easier for future changes, like supporting additional `fareTypes` as it's not reliant on ordering, so won't result in accidental reordering of the lists resulting in incorrect messages appearing in the rendered component

repos/fdbt-site/src/pages/ticketRepresentation.tsx
```diff
@@ -37,6 +37,26 @@ const getFareTypeDesc = (fareType: TicketType) => {
...
+        case 'flatFare':
+            return [
+                'Single trip anywhere within a geographic zone',
+                'Single trip on a specific service or set of services',
+            ];
+        default:
+            return [
+                'Unlimited travel within a geographical zone',
+                'Unlimited travel on specific service or set of services',
+            ];
```
joeds13, line 56: We should look to remove the duplication of these options, that appear both in the `period` and `default` cases, we have two options:
* If these sentences are truly generic then we should just remove the `period` case as the component will just use the `default` case anyway. However the use of the `Unlimited travel` wording is the actual cause of this bug code change, so it being the default for all `fareTypes` might lead to this same issue re-occurring in the future.
* We change these `default` responses to be truly generic, this means that if a `fareType` that's not `period` or `flatFare` gets to this page (it's included in a new future features journey) then the wording seen is initially suitable, and could have it's own case created if needed.

joeds13 (changes requested):

repos/fdbt-site/src/pages/ticketRepresentation.tsx
Aaron-Collis-IW: This is also done

joeds13 (approved):

PR #514, opened 2022-09-12
feat: edit product name for period tickets

RobCattonKPMG:
## Description

To allow users to edit period tickets' product names.

## Testing instructions

Run locally, create a period product, and then in My Fares attempt to edit it via the edit link. Point to point products should NOT be able to have their names edited.


NitishAroraKPMG (approved):
Looks fine

PR #515, opened 2022-09-12
feat: updated site to allow editing point to point product names

RobCattonKPMG:
## Description

Updated site to allow editing point to point product names changed exporter to give product names if not present, altered the netex converter to use product names from singles and returns instead of the combination of passenger type and faretype.

## Testing instructions

Use site locally to edit a point to point ticket name, use Makefile in dev repo to ensure valid netex is generated.


NitishAroraKPMG (approved):
looks file

PR #516, opened 2022-09-12
fix: Changed libxslt version back from 10.0.1 to 10.0.0

RobCattonKPMG:
## Description

To fix the netex converter, which currently fails with an error saying: lib64/libm.so.6: version `GLIBC_2.29' not found (required by /var/task/node_modules/node1-libxmljsmt-myh/build/Release/xmljs-myh.node)


NitishAroraKPMG (approved):
Dependency changes looks good to me

PR #519, opened 2022-09-13
feat: duplicate products

RobCattonKPMG:
## Description

To give users the ability to duplicate a product, so that they can quickly create new tickets if they are very similar.

## Testing instructions

Create a product and then hit copy in the product list inside My Fares.


NitishAroraKPMG (approved):
looks fine

PR #538, opened 2022-09-27
Changes for the service

NitishAroraKPMG:
## Description

Show the service to user while selecting the return leg for circular service

## Testing instructions

Create a return ticket and select the circular service which will redirect the user to select the another service to be used in inbound matching


repos/fdbt-site/src/pages/returnService.tsx
```diff
@@ -149,7 +149,7 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
                 const outboundDirection = directions.find((it) => ['outbound', 'clockwise'].includes(it));
                 const inboundDirection = directions.find((it) => ['inbound', 'antiClockwise'].includes(it));
 
-                if (!outboundDirection || !inboundDirection) {
+                if (inboundDirection || outboundDirection) {
```
RobCatton, line 152: Why have we changed this to check if either is present? This means that we are returning any service with 1 or 2 directions - we should only be returning services with 1 direction.

NitishAroraKPMG: reverted to the original state

SamerIssam (approved):

PR #561, opened 2022-10-11
feat: updated name of fare table fare frame

RobCattonKPMG:
## Description

To update the name of the fare table fare frame, to be 'Prices for <product name>'.

## Testing instructions

None needed - plain text change to XML so no risk of validation fail.


NitishAroraKPMG (approved):
looks good

PR #562, opened 2022-10-11
chore: node 14 upgrade for netex converter

RobCattonKPMG:
## Description

Node 12 runtime is no longer supported, so we need to update the netex converter to use Node 14 (which is the runtime used by all other node lambdas).

## Testing instructions

Ensure all functionality of the netex converter is still there (creating tickets, exporting).


SamerIssam (approved):
Looks good.
Tested locally. 
I was able to install and run `make generate-validate-all` with no issues

PR #563, opened 2022-10-12
fix: fixing tariff name

RobCattonKPMG:
## Description

Tariff name wasnt being properly altered - it was only happening inside the Preassigned Fare Product.

## Testing instructions

Generate some netex, and observe that the name of the Tariff is 'Tariff for <product-name>' instead of only '<product-name>'.


SamerIssam (approved):
Looks good

PR #564, opened 2022-10-12
feat: Adding UserType to limitations NetEx

alainajaffer:
## Description

To add UserType to limitations where the useType is one of the valid passengerTypes as shown on the site.

## Testing instructions

Check limitations/UserProfile and find UserType should be the same value as name

### DIFF
--- a/repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
+++ b/repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
@@ -99,6 +99,7 @@ export const getUserProfile = (user: User | GroupCompanion, index: number): Nete
     version: '1.0',
     id: `op:${user.passengerType}-${index}`,
     Name: { $t: user.passengerType },
+    UserType: { $t: user.passengerType },
     TypeOfConcessionRef: {
         version: 'fxc:v1.0',
         ref: `fxc:${
--- a/repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
+++ b/repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
@@ -26,6 +26,7 @@ import {
     isHybridTicket,
     isReturnTicket,
     isSingleTicket,
+    isPassengerType,
 } from '../types';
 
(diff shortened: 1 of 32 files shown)


repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
```diff
@@ -95,20 +96,22 @@ export const getProfileRef = (
...
-        ref: `fxc:${
-            user.passengerType === 'anyone' || user.passengerType === 'adult' ? 'none' : snakeCase(user.passengerType)
-        }`,
-    },
-    MinimumAge: { $t: user.ageRangeMin || null },
-    MaximumAge: { $t: user.ageRangeMax || null },
-    ProofRequired: { $t: user.proofDocuments?.join(' ') || null },
-});
+export const getUserProfile = (user: User | GroupCompanion, index: number): NetexObject => {
+    const passengerType = isPassengerType(user.passengerType) ? user.passengerType : '';
```
RobCattonKPMG, line 100: I think we need to be throwing an error here if the passenger type is not one of the accepted ones, because something has gone very very wrong if so, and we need to be alerted to it.

RobCattonKPMG: so something like 

`if (!isPassengerType(user.passengerType)) { throw new Error('Passenger type received was not one of the accepted enumerations.')`

RobCattonKPMG: Alternatively you could move all of this into a function, and have that function be the one to throw the error, to make this function a little bit more concise. If you do this, which I think makes sense, we could execute the function at the start of the netex generator, not all the way down here.

alainajaffer: [476457a](https://github.com/fares-data-build-tool/create-data/pull/564/commits/476457a1a23051d392667a04692cd604b61afa5b)

RobCattonKPMG (approved):

PR #566, opened 2022-10-12
fix: product type bug and return validity default set

RobCattonKPMG:
## Description

This PR does a few things. There was a bug in the Netex-converter where we were setting a value based on the ticket period being in days, when it should have also been checking that the number of days was '1'. Also, we are setting the return period validity to 1 day, if the user selects no on the returnValidity page in the site.

Lastly, this PR alters our test data so that all of the tickets we use have 1 product in, which is inline with what the site generates.

## Testing instructions

Generate some Netex and view that <ProductType> correctly is set for tickets that are 1 day (they should be dayPasses etc).


alainajaffer:
multiOpGeoZone.xml has a name of 1 day but it product type period pass is that intended?

alainajaffer (approved):

PR #569, opened 2022-10-13
fix: correct metadata error

SamerIssam:
## Description

Correcting metadata error 

## Testing instructions
generated netex files has no senior - child error/


NitishAroraKPMG (approved):
looks fine

PR #574, opened 2022-10-17
fix: Moving env checks to server side props

RobCattonKPMG:
## Description

To turn on multi operator for test env only.

## Testing instructions

Locally it works, needs checking on test environment.


SamerIssam (approved):
looks good

PR #576, opened 2022-10-18
feat: Altered netex file naming strategy 

RobCattonKPMG:
## Description

We have changed the way we name our generated netex files, instead going for a structure like this:

FX-PI-01_UK_MCTR_LINE-FARE_Best-Product_2022-10-18_2010-12-17.xml

And we add _1 or _2 etc if there are duplicates.

## Testing instructions

Locally, this code will not be ran - to ensure the make commands all still work (it would have to be edited to make it work with this change). Unit test coverage ensures that the path where it was saving the files should remain the same. There is some uncertainty around how this will affect scheme operators generating netex, but the current site doesn't support them fully anyway.

### DIFF
--- a/repos/fdbt-netex-output/src/netex-convertor/handler.ts
+++ b/repos/fdbt-netex-output/src/netex-convertor/handler.ts
@@ -6,6 +6,9 @@ import * as s3 from '../data/s3';
 import { isSchemeOperatorTicket, Ticket } from '../types/index';
 import netexGenerator from './netexGenerator';
 import { Operator } from '../../src/types/index';
+import { getProductType } from './sharedHelpers';
+import { SchemeOperatorTicket } from 'fdbt-types/matchingJsonTypes';
+import { fileNameExistsAlready } from '../data/s3';
 
 export const xsl = `
     <xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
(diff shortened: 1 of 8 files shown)


repos/fdbt-netex-output/src/netex-convertor/handler.ts
```diff
@@ -45,7 +48,53 @@ const uploadToS3 = async (netex: string, fileName: string): Promise<void> => {
...
+    const startDate = ticket.ticketPeriod.startDate.split('T')[0];
+    return `FX-PI-01_UK_${nocOrSchemeName}_${lineOrNetworkFare}_${productName}_${creationDate}_${startDate}`;
+};
+
+export const checkToSeeIfFileNameExists = async (fileName: string): Promise<string> => {
+    let placeHolderFileName = fileName;
+    let cannotProceed = true;
+    let counter = 1;
+
+    while (cannotProceed) {
```
NitishAroraKPMG, line 82: we can improve the code 

while (true) {

}

repos/fdbt-netex-output/src/netex-convertor/handler.ts
```diff
@@ -45,7 +48,53 @@ const uploadToS3 = async (netex: string, fileName: string): Promise<void> => {
...
+        const fileNameTaken = await fileNameExistsAlready(placeHolderFileName);
+        if (fileNameTaken) {
+            if (counter > 1) {
+                placeHolderFileName = `${placeHolderFileName.substring(0, placeHolderFileName.length - 2)}_${counter}`;
+            } else {
+                placeHolderFileName = `${placeHolderFileName}_${counter}`;
+            }
+            counter++;
+        } else {
+            cannotProceed = false;
```
NitishAroraKPMG, line 92: we can break the loop here using 
break;

repos/fdbt-netex-output/src/netex-convertor/handler.ts
```diff
@@ -45,7 +48,53 @@ const uploadToS3 = async (netex: string, fileName: string): Promise<void> => {
...
+export const checkToSeeIfFileNameExists = async (fileName: string): Promise<string> => {
+    let placeHolderFileName = fileName;
+    let cannotProceed = true;
+    let counter = 1;
+
+    while (cannotProceed) {
+        const fileNameTaken = await fileNameExistsAlready(placeHolderFileName);
+        if (fileNameTaken) {
+            if (counter > 1) {
+                placeHolderFileName = `${placeHolderFileName.substring(0, placeHolderFileName.length - 2)}_${counter}`;
```
NitishAroraKPMG, line 86: can we do it like this:

placeHolderFileName = `${fileName}_${counter}`;

repos/fdbt-netex-output/src/netex-convertor/handler.ts
```diff
@@ -45,7 +48,53 @@ const uploadToS3 = async (netex: string, fileName: string): Promise<void> => {
...
+    const productType = getProductType(ticket);
+    const lineOrNetworkFare = getLineOrNetworkFare(productType);
+    const nocOrSchemeName = isSchemeOperatorTicket(ticket) ? getSchemeNocIdentifier(ticket) : ticket.nocCode;
+    const productName = ticket.products[0].productName.replace(' ', '-');
+    const creationDate = new Date(Date.now()).toISOString().split('T')[0];
+    const startDate = ticket.ticketPeriod.startDate.split('T')[0];
+    return `FX-PI-01_UK_${nocOrSchemeName}_${lineOrNetworkFare}_${productName}_${creationDate}_${startDate}`;
+};
+
+export const checkToSeeIfFileNameExists = async (fileName: string): Promise<string> => {
```
NitishAroraKPMG, line 77: we are not checking here, can we add/rename like
getFileNameIfExists or just getNetexFileName

repos/fdbt-netex-output/src/netex-convertor/handler.ts
RobCattonKPMG: Good call!

repos/fdbt-netex-output/src/netex-convertor/handler.ts
RobCattonKPMG: Will do

repos/fdbt-netex-output/src/netex-convertor/handler.ts
RobCattonKPMG: The problem if we do that, is that if it fails the check again, we will end up with a name like `<fileName>_1_2`

repos/fdbt-netex-output/src/netex-convertor/handler.ts
RobCattonKPMG: Yep will rename.

NitishAroraKPMG (dismissed):
looks good

NitishAroraKPMG (approved):
looks fine

PR #578, opened 2022-10-18
fix: Passenger breaks product

alainajaffer:
## Description

When you delete a passengerType that is connected to a product it will cause an error. To resolve this I have added an error message if the user attempts to delete a passenger that has products. The checks added will also prevent them from deleting the passengerType if there are connected products.

## Testing instructions

**Case 1:**
Make a product and connect it to a passenger type.
Try delete the passenger type.
An error message on the page should appear.

_All valid examples:_
You cannot delete 123 because it is being used in 3 products.

**Case 2:**
Make a passenger type alone that is not connected to a product.
Try deleting this.
It should be removed.

Note to add a passenger navigate to View and manage fares > Operator settings > Passenger types

### DIFF
--- a/repos/fdbt-site/fdbt-types/matchingJsonTypes.ts
+++ b/repos/fdbt-site/fdbt-types/matchingJsonTypes.ts
@@ -1,5 +1,4 @@
 export interface PointToPointCarnetProductDetails extends BaseProduct {
-    productName: string;
     carnetDetails: CarnetDetails;
 }
 
@@ -33,7 +32,6 @@ export interface PeriodMultipleServicesTicket extends BasePeriodTicket {
 }
 
 export interface FlatFareProductDetails extends BaseProduct {
-    productName: string;
     productPrice: string;
     carnetDetails?: CarnetDetails;
 }
@@ -240,7 +238,6 @@ export interface Product {
 }
 
 export interface ProductDetails extends BaseProduct {
-    productName: string;
     productPrice: string;
     productDuration: string;
     productValidity: string;
@@ -273,6 +270,7 @@ export interface TicketPeriod {
 }
 
 export interface BaseProduct {
+    productName: string;
     salesOfferPackages: SalesOfferPackage[];
 }
 
--- a/repos/fdbt-site/fdbt-types/matchingJsonTypes.ts
+++ b/repos/fdbt-site/fdbt-types/matchingJsonTypes.ts
@@ -1,4 +1,5 @@
 export interface PointToPointCarnetProductDetails extends BaseProduct {
+    productName: string;
     carnetDetails: CarnetDetails;
 }
 
@@ -32,6 +33,7 @@ export interface PeriodMultipleServicesTicket extends BasePeriodTicket {
 }
 
 export interface FlatFareProductDetails extends BaseProduct {
+    productName: string;
     productPrice: string;
     carnetDetails?: CarnetDetails;
 }
@@ -238,6 +240,7 @@ export interface Product {
 }
 
 export interface ProductDetails extends BaseProduct {
+    productName: string;
     productPrice: string;
     productDuration: string;
     productValidity: string;
@@ -270,7 +273,6 @@ export interface TicketPeriod {
 }
 
 export interface BaseProduct {
-    productName: string;
     salesOfferPackages: SalesOfferPackage[];
 }
 
--- a/repos/fdbt-site/src/pages/api/deletePassenger.ts
+++ b/repos/fdbt-site/src/pages/api/deletePassenger.ts
@@ -1,10 +1,15 @@
 import { NextApiResponse } from 'next';
-import { deletePassengerTypeByNocCodeAndId, getAllProductsByNoc, getGroupPassengerTypeDbsFromGlobalSettings, getPassengerTypeById } from '../../data/auroradb';
+import {
+    deletePassengerTypeByNocCodeAndId,
+    getAllProductsByNoc,
+    getGroupPassengerTypeDbsFromGlobalSettings,
+    getPassengerTypeById,
+} from '../../data/auroradb';
 import { redirectToError, redirectTo, getAndValidateNoc } from '../../utils/apiUtils/index';
 import { ErrorInfo, NextApiRequestWithSession } from '../../interfaces';
 import { getProductsMatchingJson } from 'src/data/s3';
 import { updateSessionAttribute } from 'src/utils/sessions';
-import { VIEW_PASSENGER_TYPE} from 'src/constants/attributes';
+import { VIEW_PASSENGER_TYPE } from 'src/constants/attributes';
 
 export default async (req: NextApiRequestWithSession, res: NextApiResponse): Promise<void> => {
     try {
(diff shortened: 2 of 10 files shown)


repos/fdbt-site/src/pages/api/deletePassenger.ts
```diff
@@ -25,6 +28,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
             throw new Error('This individual cannot be deleted because it is in use in a group.');
         }
 
+        const products = await getAllProductsByNoc(nationalOperatorCode)
+        const matchingJsonLinks = products.map((product) => product.matchingJsonLink);
+        const tickets = await Promise.all(
+            matchingJsonLinks?.map(async (link) => {
```
NitishAroraKPMG, line 34: do we require it like ?.
 we can use matchingJsonLinks.map instead

repos/fdbt-site/src/pages/api/deletePassenger.ts
```diff
@@ -25,6 +28,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+                return await getProductsMatchingJson(link);
+            }),
+        );
+
+        const productsUsingPassengerType = tickets?.filter(t => t?.passengerType?.id === id)
```
NitishAroraKPMG, line 39: remove ?. in the lines also

repos/fdbt-site/src/pages/api/deletePassenger.ts
```diff
@@ -25,6 +28,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+        const productNames = productsUsingPassengerType?.map(ticket=> ticket.products[0] ? (ticket.products[0].productName || 'missing'): "")
+        if(productsUsingPassengerType && productsUsingPassengerType.length > 0){
+            const passengerDetails : any = await getPassengerTypeById(id, nationalOperatorCode)
+            const { name } = passengerDetails || ""
```
NitishAroraKPMG, line 43: is blank passenger type allowed?

repos/fdbt-site/src/pages/api/deletePassenger.ts
```diff
@@ -25,6 +28,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+        const tickets = await Promise.all(
+            matchingJsonLinks?.map(async (link) => {
+                return await getProductsMatchingJson(link);
+            }),
+        );
+
+        const productsUsingPassengerType = tickets?.filter(t => t?.passengerType?.id === id)
+        const productNames = productsUsingPassengerType?.map(ticket=> ticket.products[0] ? (ticket.products[0].productName || 'missing'): "")
+        if(productsUsingPassengerType && productsUsingPassengerType.length > 0){
+            const passengerDetails : any = await getPassengerTypeById(id, nationalOperatorCode)
```
NitishAroraKPMG, line 42: const passengerType  =  await getPassengerTypeById(id, nationalOperatorCode)

repos/fdbt-site/src/pages/viewPassengerTypes.tsx
```diff
@@ -19,21 +20,23 @@ interface PassengerTypeProps {
     singlePassengerTypes: SinglePassengerType[];
     groupPassengerTypes: FullGroupPassengerType[];
     referer: string | null;
+    productsUsingPassengerType: { errors: ErrorInfo[]; productsUsingPassengerType: string[]; passengerName: string; } | null;
```
NitishAroraKPMG, line 23: we are only using the errors, not the other attributes like productsUsingPassengerType and passengerName, remove them

repos/fdbt-site/src/pages/api/deletePassenger.ts
```diff
@@ -25,6 +28,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+            const { name } = passengerDetails || ""
+            const errorMessage = `You cannot delete ${name} because ${productNames.some(p => p!=='missing') ? `it is part of the following product(s): ${productNames?.join(', ')}.`: 'it is in use in a product(s).'} `
+            const errors: ErrorInfo[] = [{ id: '/viewPassengerTypes', errorMessage}];
+            updateSessionAttribute(req, PRODUCTS_USING_PASSENGER_TYPE, {
```
NitishAroraKPMG, line 46: instead add the session attribute for the errors ... 
VIEW_PASSENGER_TYPE

and assign the errors in that and use the same in while displaying it

NitishAroraKPMG (commented):
please check

repos/fdbt-site/src/pages/api/deletePassenger.ts
alainajaffer: if matchingJson comes out null .map will throw an error ?. prevents that

repos/fdbt-site/src/pages/api/deletePassenger.ts
alainajaffer: same as above

repos/fdbt-site/src/pages/api/deletePassenger.ts
alainajaffer: throws error as it could be null so this prevents that

repos/fdbt-site/src/pages/api/deletePassenger.ts
alainajaffer: how would i prevent the typescript error here it will throw one without it?

repos/fdbt-site/src/pages/viewPassengerTypes.tsx
alainajaffer: sure working on it

repos/fdbt-site/src/pages/api/deletePassenger.ts
alainajaffer: [c54320c](https://github.com/fares-data-build-tool/create-data/pull/578/commits/c54320c83560a526531b3cd112f6764e27f23dfe)

repos/fdbt-site/src/pages/viewPassengerTypes.tsx
alainajaffer: [c54320c](https://github.com/fares-data-build-tool/create-data/pull/578/commits/c54320c83560a526531b3cd112f6764e27f23dfe)

repos/fdbt-site/fdbt-types/matchingJsonTypes.ts
```diff
@@ -273,6 +270,7 @@ export interface TicketPeriod {
 }
 
 export interface BaseProduct {
```
RobCattonKPMG, line 272: I think we need to put these types back to how they were, because productName is only *always* inside the product after the ticket has gone through the exporter - in the site, it might not have a product name. Sorry for misleading you on this!

alainajaffer: [1eb22e6](https://github.com/fares-data-build-tool/create-data/pull/578/commits/1eb22e63ffd9b58ee969e33b1648e09d7e69f501)

RobCattonKPMG (dismissed):

RobCattonKPMG (approved):

PR #581, opened 2022-10-19
fix: Passenger type card ID's

RobCattonKPMG:
## Description

Giving passenger type cards ID's so they can be referenced in errors.

## Testing instructions

Attempt to delete a passenger type that is being used by a product, and then click on the red error you are shown. It should snap the user's focus to the passenger type list.


NitishAroraKPMG (approved):
looks fine

PR #586, opened 2022-10-20
fix: Fixed netex file name creation

RobCattonKPMG:
## Description

There was a bug in creating the names for netex files, where it wasn't checking for them properly. We have also added lineName and direction for point to point tickets, to make it less likely there are duplicates.

## Testing instructions

Needs running on test environment to do a proper export.


alainajaffer:
Replace `singleTicket.ticketPeriod.startDate.split('T')[0]` with `⏎················singleTicket.ticketPeriod.startDate.split('T')[0]⏎············`

alainajaffer (approved):

PR #591, opened 2022-10-24
fix: Deleting products now deletes in S3

RobCattonKPMG:
## Description

Previously when a product was being deleted, it was only being deleted from the DB and the s3 product was being left behind. 

## Testing instructions

Delete a product, and then check the s3 bucket to see if it was removed from there also.


SamerIssam (approved):
looks good

PR #593, opened 2022-10-24
fix: Removing deletion for matching bucket

RobCattonKPMG:
## Description

We cannot delete from the matching bucket for a product as we dont know the name of the export associated.

## Testing instructions

N/A


SamerIssam (approved):
looks good

PR #594, opened 2022-10-24
Preprod Release

RobCattonKPMG:
feat: Altered netex file naming strategy
feat: Include check for BODS services after operator search
feat: Created Multioperator tab in My Fares
feat: added operator group in operator settings

fix: Stop users deleting time restrictions used in products
fix: Stop users deleting purchase methods used in products
fix: Stop users deleting passenger types used in products
fix: Adding passenger type card IDs to UI
fix: Changed version of aws amazon ecs deploy task
fix: Deleting products now deletes in S3 products bucket

test: Disabling scheme UI tests and enabling multioperator ones

chore: Added linting rules to types repo


SamerIssam (approved):
looks good

PR #595, opened 2022-10-25
fix: Changed GS operator groups so it is only enabled on test

RobCattonKPMG:
## Description

We need to ensure Global Settings operator groups is not enabled in prod, because it is only half finished.

## Testing instructions

Check operator groups are visible on test, but not on preprod or prod.


SamerIssam (approved):
looks good

SamerIssam (approved):
looks good

PR #598, opened 2022-10-25
fix: cancel exports fix

RobCattonKPMG:
## Description

Cancel exports button not working in test, attempted fix and added extra logging.

## Testing instructions

Cancel an export on test.


SamerIssam (approved):
Looks good

PR #600, opened 2022-10-26
Operator group selection is mandatory

NitishAroraKPMG:
## Description

User has to either create a new operator group or select from the already existing operator group for multi operator ticket

## Testing instructions

Create a multi operator ticket.


repos/fdbt-site/src/pages/api/reuseOperatorGroup.ts
```diff
@@ -12,34 +12,27 @@ import { updateSessionAttribute, getSessionAttribute } from '../../utils/session
...
-        }
+        const multipleOperators = await getOperatorGroupByNocAndId(selectedOperatorGroup, noc);
 
-        if (!premadeOperatorGroup) {
-            updateSessionAttribute(req, REUSE_OPERATOR_GROUP_ATTRIBUTE, [
-                { errorMessage: 'Choose a premade operator group from the options below', id: 'premadeOperatorGroup' },
-            ]);
-            redirectTo(res, '/reuseOperatorGroup');
-            return;
+        if (multipleOperators) {
```
RobCattonKPMG, line 28: We are missing an 'else' on the end of this if statement. If the user modifies the form and changes their ID, and our DB call returns undefined, we need to be throwing an error instead of letting the code continue.

alainajaffer:
Tested and seems to work just waiting on @NitishAroraKPMG  to make some ui cosmetic changes and then to resolve comments above

repos/fdbt-site/src/pages/api/reuseOperatorGroup.ts
NitishAroraKPMG: Added the new case for the else and write a test case for the same.

alainajaffer (approved):

PR #601, opened 2022-10-26
feat: cancel exports button

RobCattonKPMG:
## Description

To allow users to cancel exports that have frozen.

## Testing instructions

Cancel an export that has frozen on test env (do not cancel while in progress).

### DIFF
--- a/repos/fdbt-site/src/utils/index.ts
+++ b/repos/fdbt-site/src/utils/index.ts
@@ -241,3 +241,9 @@ export const convertDateFormat = (date: string): string => {
 };
 
 export const isReturnTicket = (ticket: Ticket | TicketWithIds): ticket is ReturnTicket => ticket.type === 'return';
+
+export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName: string): boolean => {
+    const parts = objectKey.split('/');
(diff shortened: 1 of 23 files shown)


repos/fdbt-site/src/utils/index.ts
```diff
@@ -241,3 +241,9 @@ export const convertDateFormat = (date: string): string => {
 };
 
 export const isReturnTicket = (ticket: Ticket | TicketWithIds): ticket is ReturnTicket => ticket.type === 'return';
+
+export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName: string): boolean => {
+    const parts = objectKey.split('/');
+    const exportNamePart = parts[2];
```
alainajaffer, line 247: should we add a try catch and throw an error if parts[2] is undefined?

NitishAroraKPMG (approved):
looks fine

repos/fdbt-site/src/utils/index.ts
RobCattonKPMG: Although I've just realised we need to be catering for objects in s3 that dont match this structure - so we cant throw an error.

PR #602, opened 2022-10-26
feat: Deduplicating services

alainajaffer:
## Description

The service list presented during the multiple services journey should be deduplicated and one service presented per line id. Date should be removed.  

## Testing instructions

Add a multi operator product and check the checklist throughout - there should be no date and each item should be unique per lineId


SamerIssam (approved):
Looks good

PR #613, opened 2022-11-04
feat: Use the operator group from operator settings 

NitishAroraKPMG:
## Description

Use the saved or already created operator group in the multi operator ticket journey.

## Testing instructions

Create the multi operator ticket and user will be asked to select the operator group if exists else needs to be created on the fly.
Edit the operator group from the operator settings 
Delete the operator group from operator settings.
Json file for the ticket will be having a operator group id to track the tickets using the operator group.


alainajaffer:
Screenshot 2022-11-04 at 12 34 05
Can we remove the white line here?

alainajaffer:
Seems to work just have a couple business questions

alainajaffer (approved):

PR #614, opened 2022-11-07
feat: Group of lines

alainajaffer:
## Description

Where multiple Lines are included in a file, these should be combined into a GroupOfLines to be referenced more efficiently elsewhere. 

## Testing instructions

Check generated netex to see if group of line where there are multiple line references 

Ticket types with multiple lines (flat fare, period multi service, etc) all need to be creating a group of lines, and then referencing it in the tariff.

### DIFF
--- a/repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
+++ b/repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
@@ -451,14 +451,24 @@ export const getFareStructuresElements = (
 
         if ('productDuration' in product) {
             return [
-                getPeriodAvailabilityElement(availabilityElementId, validityParametersObject, hasTimeRestriction, groupOfLinesRef),
+                getPeriodAvailabilityElement(
+                    availabilityElementId,
+                    validityParametersObject,
+                    hasTimeRestriction,
+                    groupOfLinesRef,
+                ),
                 getDurationElement(ticket, product),
                 getPeriodConditionsElement(ticket, product),
             ];
         }
 
         return [
-            getPeriodAvailabilityElement(availabilityElementId, validityParametersObject, hasTimeRestriction, groupOfLinesRef),
+            getPeriodAvailabilityElement(
+                availabilityElementId,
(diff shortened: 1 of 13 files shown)


repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
```diff
@@ -449,14 +451,24 @@ export const getFareStructuresElements = (
 
         if ('productDuration' in product) {
             return [
-                getPeriodAvailabilityElement(availabilityElementId, validityParametersObject, hasTimeRestriction),
+                getPeriodAvailabilityElement(
+                    availabilityElementId,
+                    validityParametersObject,
+                    hasTimeRestriction,
+                    groupOfLinesRef,
```
NitishAroraKPMG, line 458: do we need to add this as a parameter?

repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
```diff
@@ -449,14 +451,24 @@ export const getFareStructuresElements = (
...
         }
 
         return [
-            getPeriodAvailabilityElement(availabilityElementId, validityParametersObject, hasTimeRestriction),
+            getPeriodAvailabilityElement(
+                availabilityElementId,
+                validityParametersObject,
+                hasTimeRestriction,
+                groupOfLinesRef,
+            ),
```
NitishAroraKPMG, line 471: do we need to change the calling of function?

alainajaffer: It is being referenced in the function so if it is present it will use this value

repos/fdbt-netex-output/src/netex-convertor/sharedHelpers.ts
alainajaffer: It is being referenced in the function so if it is present it will use this value

NitishAroraKPMG (approved):
looks fine

PR #624, opened 2022-11-10
feat: Edit operator group services

RobCattonKPMG:
## Description

Feature to allow the user to edit the services they have selected for their operator group. Also some refactoring and logging added.

## Testing instructions

Create an operator group, and use it to make a multi operator ticket. Then edit the services you selected for the additional operators, and ensure it has updated. Also, test around the operator group area in general as there was a lot of refactoring.


repos/fdbt-netex-output/src/netex-convertor/handler.test.ts
```diff
@@ -377,7 +377,7 @@ describe('netexConvertorHandler', () => {
             await netexConvertorHandler(event);
         } catch {
             const expectedObject = {
-                Message: 'There was an error when converting the NeTEx file: ',
+                Message: 'There was an error when converting the NeTEx file: MCTR/exports/MCTR_2022_09_30/MCTRd00f5a99_1664555550706.json',
```
alainajaffer, line 380: lint issue

alainajaffer:
if I try to change LNUD additional services it will take me to the /multipleOperatorsServiceList page but it will not pre populate the services I already have - do we want this to happen? Technically as a user i may have forgotten what services I had for those additional ones

alainajaffer:
I assumed we were maintaining the operator group name in the query string of /searchOperators. So when I trigger a new search the name would still be there.

alainajaffer:
When I edit an operator out of the operator group and go to edit the mutli operator product the one i removed it still there - what do we expect in this case?

alainajaffer (approved):
Approved as comments to be addressed in another ticket

PR #643, opened 2022-11-17
feat: select products to export

RobCattonKPMG:
## Description

Allowing a user to select which products they want to export, via checkboxes. 

Also adding in more date validation, so start / end dates beyond 2099 cannot be used.

## Testing instructions

Actual export will only run properly on test environment, but all other functionality can be done locally.


repos/fdbt-site/src/pages/api/selectExports.ts
```diff
@@ -0,0 +1,46 @@
+import { NextApiResponse } from 'next';
+import { getAndValidateNoc, redirectTo } from '../../utils/apiUtils';
+import { NextApiRequestWithSession } from '../../interfaces';
+import { getS3Exports } from '../../data/s3';
+import { getProductById } from '../../data/auroradb';
+import { triggerExport } from '../../utils/apiUtils/export';
+
+export default async (req: NextApiRequestWithSession, res: NextApiResponse): Promise<void> => {
+    const noc = getAndValidateNoc(req, res);
+    const productIdsFromReq: string[] | undefined | string = req.body.productsToExport;
```
alainajaffer, line 10: could req.body.productsToExport this line fail?
also why is there three types it could be

RobCattonKPMG: It wont throw an error, it will just be undefined if the user doesn't send anything across in the body. It can be three types because if the user selects one checkbox, the body will look like this 

{
productsToExport: '1'
}

But if there are two, because they both use the same 'name' in the input, it sends them across as an array, like this:

{
productsToExport: [ '1', '2' ]
}

There are examples in the test file also.

alainajaffer: ok makes sense

alainajaffer (approved):

PR #653, opened 2022-11-22
fix: Removed isDevOrTest condition in the exports page

RobCattonKPMG:
## Description

As per title, removing isDevOrTest condition for the exports page, so the cancel exports button will show.

## Testing instructions

Cancel exports button should show on preprod and prod for failed exports.


NitishAroraKPMG (approved):
looks fine

PR #664, opened 2022-11-23
feat: Adding the operator settings for operator in dev environment

NitishAroraKPMG:
## Description

Creating the operator settings for the development environment

## Testing instructions

Create the local environment and run it using `make run`

### DIFF
--- a/fdbt-dev/sql/operatorGroup.sql
+++ b/fdbt-dev/sql/operatorGroup.sql
@@ -5,7 +5,7 @@ TRUNCATE TABLE `operatorGroup`;
 
 INSERT INTO `operatorGroup` (name, contents, nocCode) 
 VALUES
-('test', '[{"nocCode":"LNUD","name":"The Blackburn Bus Company"},{"nocCode":"NWBT","name": "Pilkingtonbus"}]', 'BLAC');
+('operator-group', '[{"nocCode":"LNUD","name":"The Blackburn Bus Company"},{"nocCode":"NWBT","name": "Pilkingtonbus"}]', 'BLAC');
 
 UNLOCK TABLES;
 /*!40014 SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS */;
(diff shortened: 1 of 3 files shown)


fdbt-dev/sql/operatorGroup.sql
```diff
@@ -0,0 +1,12 @@
+/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
+LOCK TABLES `operatorGroup` WRITE;
+
+TRUNCATE TABLE `operatorGroup`;
+
+INSERT INTO `operatorGroup` (name, contents, nocCode) 
+VALUES
+('test', '[{"nocCode":"LNUD","name":"The Blackburn Bus Company"},{"nocCode":"NWBT","name": "Pilkingtonbus"}]', 'BLAC');
```
alainajaffer, line 8: do we want to add this with the name test if the tests add one with this name?

fdbt-dev/sql/passengerTypes.sql
```diff
@@ -0,0 +1,14 @@
+/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
+LOCK TABLES `passengerType` WRITE;
+
+TRUNCATE TABLE `passengerType`;
+
+INSERT INTO `passengerType` (name,contents,isGroup,nocCode) 
+VALUES
+('Small People', '{"id":"","passengerType":"child","ageRangeMin":"","ageRangeMax":"18"}', 0, 'BLAC'),
```
RobCattonKPMG, line 8: Why are we leaving ID from the Insert and from the contents?

fdbt-dev/sql/operatorGroup.sql
NitishAroraKPMG: let me change to operator-group

fdbt-dev/sql/passengerTypes.sql
NitishAroraKPMG: id we can't insert as this is auto increment and the contents contain the data in this format ...

NitishAroraKPMG: i created the passenger type from the UI and use that one ... I have to check if we delete the id from single passenger type then it should not break

RobCattonKPMG (approved):

PR #666, opened 2022-11-28
refactor: Improve performance of exports pages

RobCattonKPMG:
## Description

Reducing number of DB queries, and other general improvements.

## Testing instructions

No changes should be evident, apart from exports pages should load quicker.


.github/workflows/fdbt-site_test_deployment.yml
```diff
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-    - 'develop'
+    - 'feature/CFDS-123_performance_improvements'
```
alainajaffer, line 4: is this expected?

RobCattonKPMG: Forgot to change it - put back now

alainajaffer (approved):

PR #673, opened 2022-12-01
Adding the edit functionality for product validity

NitishAroraKPMG:
## Description

Add the edit functionality for the products page (other products page)

## Testing instructions

Edit the product validity of a product listed in "otherProducts" page.

### DIFF
--- a/repos/fdbt-site/src/pages/api/periodValidity.ts
+++ b/repos/fdbt-site/src/pages/api/periodValidity.ts
@@ -1,25 +1,33 @@
 import { NextApiResponse } from 'next';
-import { updateSessionAttribute } from '../../utils/sessions';
-import { PERIOD_EXPIRY_ATTRIBUTE } from '../../constants/attributes';
+import { getSessionAttribute, updateSessionAttribute } from '../../utils/sessions';
+import {
+    MATCHING_JSON_ATTRIBUTE,
+    MATCHING_JSON_META_DATA_ATTRIBUTE,
+    PERIOD_EXPIRY_ATTRIBUTE,
+} from '../../constants/attributes';
 import { redirectToError, redirectTo, getAndValidateNoc } from '../../utils/apiUtils';
 import { ErrorInfo, NextApiRequestWithSession } from '../../interfaces';
 import { getFareDayEnd } from '../../data/auroradb';
-import { PeriodExpiry } from '../../interfaces/matchingJsonTypes';
+import { PeriodExpiry, Ticket, WithIds } from '../../interfaces/matchingJsonTypes';
+import { putUserDataInProductsBucketWithFilePath } from '../../utils/apiUtils/userData';
 
 export default async (req: NextApiRequestWithSession, res: NextApiResponse): Promise<void> => {
     try {
         const errors: ErrorInfo[] = [];
         if (req.body.periodValid) {
             const { periodValid } = req.body;
-            let { productEndTime } = req.body;
+
+            let productEndTime = '';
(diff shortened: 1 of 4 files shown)


repos/fdbt-site/src/pages/api/periodValidity.ts
```diff
@@ -1,25 +1,33 @@
...
+import { putUserDataInProductsBucketWithFilePath } from '../../utils/apiUtils/userData';
 
 export default async (req: NextApiRequestWithSession, res: NextApiResponse): Promise<void> => {
     try {
         const errors: ErrorInfo[] = [];
         if (req.body.periodValid) {
             const { periodValid } = req.body;
-            let { productEndTime } = req.body;
+
+            let productEndTime = '';
```
RobCattonKPMG, line 20: Why are we doing this? Doesn't this mean we aren't taking the value the user entered?

repos/fdbt-site/src/pages/api/periodValidity.ts
```diff
@@ -29,13 +37,34 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
                     productEndTime = endOfFareDay;
                 }
-            } else {
-                productEndTime = '';
+            }
+
+            // redirected from the product details page
+            if (ticket && matchingJsonMetaData) {
+                const product = ticket.products[0];
+                const updatedProduct = { ...product, productValidity: periodValid, productEndTime: productEndTime };
```
RobCattonKPMG, line 45: productEndTime will always be ''

repos/fdbt-site/tests/pages/api/periodValidity.test.ts
```diff
@@ -36,6 +43,43 @@ describe('periodValidity', () => {
...
+                [MATCHING_JSON_ATTRIBUTE]: expectedPeriodGeoZoneTicketWithMultipleProducts,
+                [MATCHING_JSON_META_DATA_ATTRIBUTE]: {
+                    productId: '2',
+                    matchingJsonLink: 'test/path',
+                },
+            },
+            mockWriteHeadFn: writeHeadMock,
+        });
+
+        await periodValidity(req, res);
```
RobCattonKPMG, line 60: I think we also need a test for when productEndTime is in the request, as we have missed a bug here.

repos/fdbt-site/src/pages/api/periodValidity.ts
NitishAroraKPMG: yes, it is not getting input from the user and taking from the DB call at later point of code.

repos/fdbt-site/src/pages/api/periodValidity.ts
NitishAroraKPMG: it will capture the value as per the DB

repos/fdbt-site/tests/pages/api/periodValidity.test.ts
NitishAroraKPMG: there will never be productEndTime in the request

RobCattonKPMG: It looks like there should be, because its an input the user has access to in the form (/selectPeriodValidity has an input as below 
Screenshot 2022-12-01 at 10 40 47

Im not sure why it doesn't work, as it doesnt ever send the value to the API, but it's irrelevant I think anyway.

RobCattonKPMG (approved):

PR #689, opened 2022-12-07
feat: Editing products should default

alainajaffer:
## Description

All edit functions in the product details page should offer default values previously defined for product 

## Testing instructions

Check services, other products and multi op pages try and edit some product attributes for different ticket types


NitishAroraKPMG (commented):
we need to change more files as there are lot of options we can edit from product details page.

Suggestion: Can we add a props similar to the type of page like inputs in ManagePassengerGroup from which we can fetch which is to be selected in edit mode.

alainajaffer:
> we need to change more files as there are lot of options we can edit from product details page.
> 
> Suggestion: Can we add a props similar to the type of page like inputs in ManagePassengerGroup from which we can fetch which is to be selected in edit mode.

done I think the best way to do this is use the ticket that was already retrieved

repos/fdbt-site/src/components/RadioConditionalInput.tsx
```diff
@@ -357,6 +357,7 @@ const renderConditionalRadioButton = (
             type="radio"
             value={radio.value}
             data-aria-controls={radio.dataAriaControls}
+            defaultChecked={radio.defaultChecked}
```
RobCattonKPMG, line 360: I dont think we should be changing this. This element is called 'uncheckedRadioInput' which means it should never be checked?

repos/fdbt-site/src/components/RadioConditionalInput.tsx
```diff
@@ -367,7 +368,7 @@ const renderConditionalRadioButton = (
             type="radio"
             value={radio.value}
             data-aria-controls={radio.dataAriaControls}
-            defaultChecked
+            defaultChecked={radio.defaultChecked}
```
RobCattonKPMG, line 371: As above,this one is *always* checked, so we presumably shouldn't be changing it so its sometimes checked?

repos/fdbt-site/src/pages/selectPeriodValidity.tsx
```diff
@@ -149,6 +157,14 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
               }`
             : '';
 
+    let fieldset: RadioConditionalInputFieldset = getFieldset(errors, endOfFareDay);
+    if (ticket) {
+        const product = ticket?.products[0] as PeriodExpiry;
```
RobCattonKPMG, line 162: do we need the ? mark? isnt there always a products array inside a ticket?

repos/fdbt-site/src/pages/selectTimeRestrictions.tsx
```diff
@@ -187,6 +187,9 @@ export const getServerSideProps = async (
     const timeRestrictions = await getTimeRestrictionByNocCode(nationalOperatorCode);
 
     const ticket = getSessionAttribute(ctx.req, MATCHING_JSON_ATTRIBUTE);
+    const selectedId =
+        getSessionAttribute(ctx.req, FULL_TIME_RESTRICTIONS_ATTRIBUTE)?.id || ticket?.timeRestriction?.id;
```
RobCattonKPMG, line 191:
```suggestion
        getSessionAttribute(ctx.req, FULL_TIME_RESTRICTIONS_ATTRIBUTE)?.id || ticket?.timeRestriction?.id || '';
```

repos/fdbt-site/src/pages/selectTimeRestrictions.tsx
```diff
@@ -196,7 +199,7 @@ export const getServerSideProps = async (
               }`
             : '';
 
-    return { props: { csrfToken, errors, timeRestrictions, selectedId, backHref } };
+    return { props: { csrfToken, errors, timeRestrictions, selectedId: selectedId || null, backHref } };
```
RobCattonKPMG, line 202:
```suggestion
    return { props: { csrfToken, errors, timeRestrictions, selectedId: selectedId, backHref } };
```

repos/fdbt-site/src/pages/selectTimeRestrictions.tsx
```diff
@@ -196,7 +199,7 @@ export const getServerSideProps = async (
               }`
             : '';
 
-    return { props: { csrfToken, errors, timeRestrictions, selectedId, backHref } };
+    return { props: { csrfToken, errors, timeRestrictions, selectedId: selectedId, backHref } };
```
RobCattonKPMG, line 202:
```suggestion
    return { props: { csrfToken, errors, timeRestrictions, selected, backHref } };
```

RobCattonKPMG (approved):

PR #700, opened 2022-12-13
feat: Edit purchase methods table

alainajaffer:
## Description

We need to edit the purchase methods DB table to have a new column 'isCapped'

## Testing instructions

Check if it is there locally and code review

### DIFF
--- a/fdbt-aws/sql/h.salesOfferPackage.sql
+++ b/fdbt-aws/sql/h.salesOfferPackage.sql
@@ -10,9 +10,11 @@ CREATE TABLE salesOfferPackage(
     `purchaseLocations` varchar(255) DEFAULT NULL,
     `paymentMethods` varchar(255) DEFAULT NULL,
     `ticketFormats` varchar(255) DEFAULT NULL,
-    `isCapped` boolean NOT NULL DEFAULT false,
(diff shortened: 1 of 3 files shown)


fdbt-aws/sql/h.salesOfferPackage.sql
```diff
@@ -10,6 +10,7 @@ CREATE TABLE salesOfferPackage(
     `purchaseLocations` varchar(255) DEFAULT NULL,
     `paymentMethods` varchar(255) DEFAULT NULL,
     `ticketFormats` varchar(255) DEFAULT NULL,
+    `isCapped` boolean NOT NULL DEFAULT false,
     INDEX idx_nocCode (nocCode),
```
RobCattonKPMG, line 14: Some requirements missed from the Jira ticket.

RobCattonKPMG: Also, could we use the ALTER way of doing this, as per other tables. It makes it easier to run the ALTER command on each environment, and also shows that we have changed the table.

alainajaffer: changes

RobCattonKPMG (approved):

PR #713, opened 2022-12-19
UI test for the capped ticket by products

NitishAroraKPMG:
## Description

Create the UI test for the capped ticket by the products

## Testing instructions

Run the cypress test for the capped ticket.

### DIFF
--- a/repos/fdbt-site/cypress_tests/cypress/e2e/capped.cy.ts
+++ b/repos/fdbt-site/cypress_tests/cypress/e2e/capped.cy.ts
@@ -43,7 +43,7 @@ describe('the capped faretype product journey', () => {
         isFinished()
     });
 
-    it.only('completes successfully for geozone products journey', () => {
+    it('completes successfully for geozone products journey', () => {
         selectFareType('cappedProduct', false);
         defineUserTypeAndTimeRestrictions();
         completeCappedGeoZonePages();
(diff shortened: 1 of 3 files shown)


repos/fdbt-site/cypress_tests/cypress/e2e/capped.cy.ts
```diff
@@ -41,4 +42,20 @@ describe('the capped faretype product journey', () => {
         completeCappedTapsJourney()
         isFinished()
     });
+
```
RobCattonKPMG, line 45: The PR here is completely fine and looks right. However, we have 6 tests here (2 per type of capped journey) and I think we could just have 3 if we instead randomly decide whether to do multi-service or geoZone each time.

Could we combine these tests so we have 1 per capped journey type, and have it 50% of the time choose geoZone and 50% of the time choose multi service please?

repos/fdbt-site/cypress_tests/cypress/e2e/capped.cy.ts
```diff
@@ -1,44 +1,41 @@
...
-    it('completes successfully for geozone distance journey', () => {
-        selectFareType('cappedProduct', false);
-        defineUserTypeAndTimeRestrictions();
-        completeCappedGeoZonePages()
-        completeCappedDistanceJourney()
-        isFinished()
-    });
 
-    it('completes successfully for multi-service distance journey', () => {
+    const isGeoZoneJourney = getRandomNumber(1, 2) === 1 ? true: false;
```
RobCattonKPMG, line 15: We need this on a 'per test' basis, because if it decides geoZone then it will do them *all* as geoZone. We want it to randomly decide in each test. Dont worry about the name of the test, the name of the test should just be which *capped* journey its taking.

RobCattonKPMG (approved):

PR #729, opened 2023-01-09
fix: netex getOperatorDataByNocCode function returns duplicates

alainajaffer:
## Description

The getOperatorDataByNocCode function in the netex repo can return the same operator more than once if they are in the NOC database more than once.

## Testing instructions

For testing purposes, create a multi operator group with 'KEVS' in the group, as they are in the NOC database (bus and coach).

Code review or test it separately / add logs


repos/fdbt-netex-output/src/data/auroradb.ts
```diff
@@ -77,7 +77,39 @@ export const getOperatorDataByNocCode = async (nocCodes: string[]): Promise<Oper
             throw new Error(`No operator data found for nocCodes: ${cleansedNocs.join(',')}`);
         }
 
-        return operatorData;
+        const allNocCodes: string[] = [];
+        let nocCount: { [key: string]: number } = {};
```
RobCattonKPMG, line 81: When we have an object like this, its always better to not use 'key', because that doesnt help the person reading it to understand what the keys of the object will be, only that they will be a string.

`{ [nocCode: string]: number }` would be more useful, because its the nocCode that we use.

repos/fdbt-netex-output/src/data/auroradb.ts
```diff
@@ -77,7 +77,39 @@ export const getOperatorDataByNocCode = async (nocCodes: string[]): Promise<Oper
...
+
+        for (let i = 0; i < operatorData.length; i++) {
```
RobCattonKPMG, line 83: Usually, we should be using a .forEach or `for (const operator of operatorData)` instead of a for loop like this.

RobCattonKPMG (approved):

PR #730, opened 2023-01-10
feat: Adding the new tnds services for modes

NitishAroraKPMG:
## Description

Enabling the operators which do not have the BODS services but the TNDS service having multi modes.

## Testing instructions

Create the different kind of tickets like single, return, period with multi services, capped ticket and the multi-operator tickets for the operator as mentioned in the description.

### DIFF
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -161,8 +161,9 @@ export const getTndsServicesByNocAndModes = async (nocCode: string, modes: strin
     const nocCodeParameter = replaceInternalNocCode(nocCode);
     logger.info('', {
         context: 'data.auroradb',
-        message: 'retrieving services for given noc',
-        noc: nocCode,
+        message: 'retrieving tnds services for given noc and modes',
+        nocCode,
+        modes,
     });
 
     try {
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -157,6 +157,37 @@ export const getServicesByNocCodeAndDataSource = async (nocCode: string, source:
     }
 };
 
+export const getTndsServicesByNocAndModes = async (nocCode: string, modes: string[]): Promise<ServiceType[]> => {
+    const nocCodeParameter = replaceInternalNocCode(nocCode);
+    logger.info('', {
+        context: 'data.auroradb',
+        message: 'retrieving services for given noc',
+        noc: nocCode,
+    });
+
+    try {
+        const substitution = modes.map(() => '?').join(',');
+        const queryInput = `
+            SELECT id, lineName, lineId, startDate, serviceDescription AS description, origin, destination, serviceCode
+            FROM txcOperatorLine
+            WHERE nocCode = ? AND dataSource = 'tnds' AND (endDate IS NULL OR CURDATE() <= endDate)
+            AND mode in (${substitution}) 
+            ORDER BY CAST(lineName AS UNSIGNED) = 0, CAST(lineName AS UNSIGNED), LEFT(lineName, 1), MID(lineName, 2), startDate;
+        `;
+
+        const queryResults = await executeQuery<ServiceType[]>(queryInput, [nocCodeParameter].concat(modes));
+
+        return (
+            queryResults.map((item) => ({
+                ...item,
+                startDate: convertDateFormat(item.startDate),
+            })) || []
+        );
+    } catch (error) {
+        throw new Error(`Could not retrieve services from AuroraDB: ${error.stack}`);
+    }
+};
+
 export const getServicesByNocCodeAndDataSourceWithGrouping = async (
     nocCode: string,
     source: string,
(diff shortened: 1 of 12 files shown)


repos/fdbt-site/src/pages/multiOperatorServiceList.tsx
```diff
@@ -466,12 +464,12 @@ export const getServerSideProps = async (
         multiOperatorData = await Promise.all(
             ticket.additionalOperators.map(async (operator) => {
                 let dataSource: 'bods' | 'tnds' = 'bods';
-                let services = await getServicesByNocCodeAndDataSourceWithGrouping(operator.nocCode, dataSource);
 
-                if (services.length === 0) {
-                    services = await getFerryAndBusServices(operator.nocCode);
+                if (modesAttribute) {
                     dataSource = 'tnds';
                 }
```
RobCattonKPMG, line 470: I dont think this is correct because if the user is logged in as a bus operator, they can no longer add Tram operators to their group. They wont have the modesAttribute in their session. I don't believe this page needs to change at all.

repos/fdbt-site/src/data/auroradb.ts
```diff
@@ -194,25 +194,26 @@ export const getServicesByNocCodeAndDataSourceWithGrouping = async (
     }
 };
 
-export const getFerryAndBusServices = async (noc: string): Promise<ServiceWithOriginAndDestination[]> => {
```
RobCattonKPMG, line 197: I think this DB call needs to go back to how it was, and a new one be created. However, the name getFerryAndBusServices needs to be getFerryAndTramServices instead.

repos/fdbt-site/src/pages/api/fareType.ts
```diff
@@ -17,15 +22,16 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
             const services = await getAllServicesByNocCode(nocCode);
             const hasBodsServices = services.some((service) => service.dataSource && service.dataSource === 'bods');
             // removed as TNDS is being disabled until further notice
-            // const hasTndsServices = services.some((service) => service.dataSource && service.dataSource === 'tnds');
+            //const hasTndsServices = services.some((service) => service.dataSource && service.dataSource === 'tnds');
+            const modes = getSessionAttribute(req, MULTI_MODAL_ATTRIBUTE);
```
RobCattonKPMG, line 26:
```suggestion
            const multiModalAttribute = getSessionAttribute(req, MULTI_MODAL_ATTRIBUTE);
```

RobCattonKPMG: if we call this modes, and we wanted to get the modes out of it, we would have to do `modes.modes` which is obviously not great.

repos/fdbt-site/src/pages/returnService.tsx
```diff
@@ -96,6 +101,7 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
 
     const ticket = getSessionAttribute(ctx.req, MATCHING_JSON_ATTRIBUTE);
     const matchingJsonMetaData = getSessionAttribute(ctx.req, MATCHING_JSON_META_DATA_ATTRIBUTE);
+    const modesAttribute = getSessionAttribute(ctx.req, MULTI_MODAL_ATTRIBUTE);
     const nocCode = getAndValidateNoc(ctx);
     const selectedServiceId = Number(ctx.query.selectedServiceId);
```
RobCattonKPMG, line 106: Do we not need to change the text on this page to mention TNDS instead of BODS? line 88

RobCattonKPMG: It should probably also get the dataSource attribute like how it is done in service.tsx line 80.

repos/fdbt-site/src/pages/serviceList.tsx
```diff
@@ -161,12 +166,13 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
...
     const serviceListAttribute = getSessionAttribute(ctx.req, SERVICE_LIST_ATTRIBUTE);
     let dataSourceAttribute = getSessionAttribute(ctx.req, TXC_SOURCE_ATTRIBUTE);
+    const modesAttribute = getSessionAttribute(ctx.req, MULTI_MODAL_ATTRIBUTE);
 
     if (!dataSourceAttribute) {
         const services = await getAllServicesByNocCode(nocCode);
         const hasBodsServices = services.some((service) => service.dataSource && service.dataSource === 'bods');
 
-        if (!hasBodsServices) {
+        if (!hasBodsServices && !modesAttribute) {
```
RobCattonKPMG, line 175: What happens if the user logs in and they have no BODS data, and they also have no modes? Wont they have a modesAttribute with an empty array?

repos/fdbt-site/src/pages/serviceList.tsx
```diff
@@ -185,7 +191,11 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
 
     const { selectAll } = ctx.query;
 
-    const chosenDataSourceServices = await getServicesByNocCodeAndDataSource(nocCode, dataSourceAttribute.source);
+    let chosenDataSourceServices;
+    if (modesAttribute) {
```
RobCattonKPMG, line 195: Do we also need to check to see if we have any modes in the array in the modesAttribute? What do we do if its empty?

repos/fdbt-site/src/pages/returnService.tsx
NitishAroraKPMG: i am using the dataSource as string instead of passing the whole attribute

repos/fdbt-site/src/pages/serviceList.tsx
NitishAroraKPMG: no, we are setting the attribute only when they have modes in it

RobCattonKPMG (approved):

repos/fdbt-site/src/data/auroradb.ts
```diff
@@ -157,6 +157,37 @@ export const getServicesByNocCodeAndDataSource = async (nocCode: string, source:
     }
 };
 
+export const getTndsServicesByNocAndModes = async (nocCode: string, modes: string[]): Promise<ServiceType[]> => {
+    const nocCodeParameter = replaceInternalNocCode(nocCode);
+    logger.info('', {
+        context: 'data.auroradb',
+        message: 'retrieving services for given noc',
```
alainajaffer, line 164: do we want to be specific that it is tnds?

PR #731, opened 2023-01-10
chore: Added new script to makefile to validate netex file directly

RobCattonKPMG:
## Description

Script to allow quicker validation of misc netex files.

## Testing instructions

Try using script with a random netex file.

### DIFF
--- a/fdbt-dev/Makefile
+++ b/fdbt-dev/Makefile
@@ -178,6 +178,10 @@ validate-netex:
 validate-latest-file:
 	./scripts/validate_latest_netex.sh
 
+# usage is 'make file=<path-to-file> validate-this-netex'
+validate-this-netex:
+	./scripts/validate_this_netex.sh $(file)
+
 copy-latest-netex:
 	EDGE_PORT=4572 awslocal s3 cp s3://fdbt-unvalidated-netex-data-dev/$(EDGE_PORT=4572 awslocal s3 ls s3://fdbt-unvalidated-netex-data-dev --recursive | sort | tail -n 1 | awk '{print $NF}') .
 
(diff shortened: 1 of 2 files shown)


fdbt-dev/Makefile
```diff
@@ -178,6 +178,10 @@ validate-netex:
 validate-latest-file:
 	./scripts/validate_latest_netex.sh
 
+# usage is 'make file=<path-to-file> validate-this-netex'
+validate-this-netex:
```
NitishAroraKPMG, line 182: can we give it good name like validate-netex-file or validate-netex

fdbt-dev/Makefile
```diff
@@ -178,6 +178,10 @@ validate-netex:
 validate-latest-file:
 	./scripts/validate_latest_netex.sh
 
+# usage is 'make file=<path-to-file> validate-this-netex'
```
NitishAroraKPMG, line 181: can we change the script name

fdbt-dev/Makefile
```diff
@@ -178,6 +178,10 @@ validate-netex:
...
+validate-this-netex:
+	./scripts/validate_this_netex.sh $(file)
```
NitishAroraKPMG, line 183: is it similar to validate-netex in the makefile

fdbt-dev/Makefile
RobCattonKPMG: We already have one named that.

fdbt-dev/Makefile
RobCattonKPMG: i dont think we need to

fdbt-dev/Makefile
RobCattonKPMG: that one doesnt work, for some unknown reason.

NitishAroraKPMG (approved):

PR #757, opened 2023-01-24
fix: Add limitation to flat fares

alainajaffer:
## Description

Fix the validation issue on the netex.

## Testing instructions

make generate validate all command


repos/fdbt-netex-output/src/netex-convertor/period-tickets/periodTicketNetexHelpers.ts
```diff
@@ -938,7 +938,7 @@ export const getPeriodConditionsElement = (
     product: { productName: string; productValidity?: string },
 ): NetexObject => {
     let usagePeriodValidity = {};
-    const type = userPeriodTicket.type;
+    const type = 'single';
```
RobCattonKPMG, line 941: Do we need this variable? I think it would be clearer if we just put the word 'single' where it is needed. We also dont need the startCase() function on it, we can just use 'Single'.

RobCattonKPMG (approved):

PR #772, opened 2023-01-30
fix: Remove nested generic parameter assignment

alainajaffer:
## Description

We have a duplicate GenericParameterAssignment being created. We need to have a separate FareStructureElement each time we reference a GroupOfLines or a FareZone. 

## Testing instructions

Which means if its a geoZone, one FareStructureElement for the zones.

If its a multiService, one FareStructureElement for the services.

If its a hyrbid, 2 fare structure elements, one for each.

If it has exempted services, one additional fare structure element.

### DIFF
--- a/fdbt-dev/data/netexData/schoolPeriodMultiService.xml
+++ b/fdbt-dev/data/netexData/schoolPeriodMultiService.xml
@@ -225,7 +225,7 @@
               </qualityStructureFactors>
               <fareStructureElements>
                 <FareStructureElement version="1.0" id="op:Tariff@Test_Product_4@access_lines">
-                  <Name>Available lines and/or zones</Name>
+                  <Name>Available zones</Name>
                   <TypeOfFareStructureElementRef version="fxc:v1.0" ref="fxc:access_when"/>
                   <qualityStructureFactors>
                     <FareDemandFactorRef ref="op@Tariff@Demand" version="1.0"/>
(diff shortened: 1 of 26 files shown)


fdbt-dev/data/netexData/schoolPeriodMultiService.xml
```diff
@@ -225,7 +225,7 @@
               </qualityStructureFactors>
               <fareStructureElements>
                 <FareStructureElement version="1.0" id="op:Tariff@Test_Product_4@access_lines">
-                  <Name>Available lines and/or zones</Name>
+                  <Name>Available zones</Name>
```
RobCattonKPMG, line 228: This says zones when its not a zonal ticket

repos/fdbt-netex-output/src/netex-convertor/period-tickets/periodTicketNetexHelpers.ts
```diff
@@ -872,12 +872,11 @@ export const getPeriodAvailabilityElement = (
...
     validityParametersObject: {},
     hasTimeRestriction: boolean,
-    productName?: string,
-    groupOfLinesRef?: string,
+    isHybrid?: boolean,
 ): NetexObject => ({
     version: '1.0',
     id: `op:${id}`,
-    Name: { $t: 'Available lines and/or zones' },
+    Name: { $t: isHybrid ? 'Available lines' : 'Available zones' },
```
RobCattonKPMG, line 879: This means it will always say Available zones, even if its not a zonal ticket

RobCattonKPMG (approved):

PR #786, opened 2023-02-01
feat: Create caps section in operator settings

alainajaffer:
## Description

Similar to passenger types etc, we need a section of operator settings where operators can create different types of caps.

This ticket is just creating the actual area in operator settings, with all the links.

## Testing instructions

Code review and navigate to operator settings and see if caps is on the sub navigation and main operator settings page.

### DIFF
--- a/repos/fdbt-site/src/interfaces/index.ts
+++ b/repos/fdbt-site/src/interfaces/index.ts
@@ -217,6 +217,7 @@ export interface SettingsOverview {
 }
 
 export interface GlobalSettingsCounts {
+    capCount: number;
     passengerTypesCount: number;
     purchaseMethodsCount: number;
     timeRestrictionsCount: number;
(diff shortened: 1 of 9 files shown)


repos/fdbt-site/src/interfaces/index.ts
```diff
@@ -269,6 +270,12 @@ export interface ResponseWithLocals extends ServerResponse {
     };
 }
 
+export interface PremadeCaps {
+    id: number;
+    contents: DbTimeRestriction[];
```
RobCattonKPMG, line 275: Why are we using DbTimeRestriction here?

RobCattonKPMG: I think because we dont know what the type will be, for now just use contents: {} and when the linting complains, just use an eslint ignore

NitishAroraKPMG (approved):

PR #788, opened 2023-02-03
feat: Adding the cap expiry in operator settings

NitishAroraKPMG:
## Description

Adding the cap expiry in operator settings

## Testing instructions

Create and update the expiry in cap

### DIFF
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -569,49 +569,6 @@ export const getAtcoCodesByNaptanCodes = async (naptanCodes: string[]): Promise<
     }
 };
 
-/**
- * For a given national operator code and line id, the function returns a
- * subset of information about each service.
- *
- *
- * @remarks
- * Only brings back services from `bods`.
- *
- *
- * @param noc - The national operator code
- * @param lineId - The line id of the service(s)
- *
- * @returns An array of RawService.
- */
-export const getServicesByNocAndLineId = async (noc: string, lineId: string): Promise<RawService[]> => {
-    const nationalOperatorCode = replaceInternalNocCode(noc);
-
-    logger.info('', {
-        context: 'data.auroradb',
-        message: 'retrieving services for a given noc and line id',
-        noc,
-        lineId,
-    });
-
-    const query = `
-      SELECT ol.id,
-             ol.startDate,
-             ol.endDate
-      FROM txcOperatorLine ol
-      WHERE ol.nocCode = ?
-      AND ol.lineId = ?
-      AND ol.dataSource = 'bods'
-    `;
-
-    const result = await executeQuery<RawService[]>(query, [nationalOperatorCode, lineId]);
-
-    return result.map((result) => ({
-        ...result,
-        startDate: convertDateFormat(result.startDate),
-        endDate: result.endDate ? convertDateFormat(result.endDate) : undefined,
-    }));
-};
-
 export const getServiceByIdAndDataSource = async (
     nocCode: string,
     id: number,
@@ -1201,37 +1158,6 @@ export const insertGroupPassengerType = async (
                          VALUES (?, ?, ?, ?)`;
 
     await executeQuery(insertQuery, [contents, true, name, nocCode]);
-};
-
-export const upsertSinglePassengerType = async (
-    nocCode: string,
-    passengerType: PassengerType,
-    name: string,
-): Promise<void> => {
-    logger.info('', {
-        context: 'data.auroradb',
-        message: 'upserting passenger type for given noc and name',
-        nocCode,
-        name,
-    });
-
-    const contents = JSON.stringify(passengerType);
-
-    try {
-        const updateQuery = `UPDATE passengerType
-                             SET contents = ?
-                             WHERE name = ?
-                             AND isGroup = ?
-                             AND nocCode = ?`;
-        const meta = await executeQuery<ResultSetHeader>(updateQuery, [contents, name, false, nocCode]);
-        if (meta.affectedRows > 1) {
-            throw Error(`Updated too many rows when updating passenger type ${meta}`);
-        } else if (meta.affectedRows === 0) {
-            await insertSinglePassengerType(nocCode, passengerType, name);
-        }
-    } catch (error) {
-        throw new Error(`Could not insert passenger type into the passengerType table. ${error}`);
-    }
 };
 
 export const updateSinglePassengerType = async (noc: string, passengerType: SinglePassengerType): Promise<void> => {
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -1725,10 +1725,10 @@ export const getCapExpiry = async (nocCode: string): Promise<string | undefined>
         `;
 
         const queryResults = await executeQuery<{ contents: string }[]>(queryInput, [nocCode]);
-        //if (queryResults.length > 0)
+
         return queryResults[0]?.contents;
     } catch (error) {
-        throw new Error(`Could not retrieve fare day end by nocCode from AuroraDB: ${error.stack}`);
+        throw new Error(`Could not retrieve cap expiry by nocCode from AuroraDB: ${error.stack}`);
     }
 };
 
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -30,7 +30,14 @@ import {
     MyFaresOtherProduct,
     DbProduct,
 } from '../interfaces/dbTypes';
-import { Stop, FromDb, SalesOfferPackage, CompanionInfo, OperatorDetails } from '../interfaces/matchingJsonTypes';
+import {
+    Stop,
+    FromDb,
+    SalesOfferPackage,
+    CompanionInfo,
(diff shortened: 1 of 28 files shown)


repos/fdbt-site/src/data/auroradb.ts
```diff
@@ -1702,6 +1709,52 @@ export const upsertFareDayEnd = async (nocCode: string, fareDayEnd: string): Pro
...
+
+    try {
+        const queryInput = `
+            SELECT contents
+            FROM caps
+            WHERE noc = ?
+            AND isExpiry = 1
+        `;
+
+        const queryResults = await executeQuery<{ contents: string }[]>(queryInput, [nocCode]);
```
RobCattonKPMG, line 1727: Will this function just throw an error if the user has no capExpiry?

repos/fdbt-site/src/pages/selectCapValidity.tsx
```diff
@@ -8,7 +8,7 @@ import { getSessionAttribute } from '../utils/sessions';
 import { getAndValidateNoc, getCsrfToken, getErrorsByIds } from '../utils';
 import RadioConditionalInput from '../components/RadioConditionalInput';
 import { isCapExpiry } from '../interfaces/typeGuards';
-import { getFareDayEnd } from '../data/auroradb';
+import { getCapExpiry, getFareDayEnd } from '../data/auroradb';
```
RobCattonKPMG, line 11: Can we add a back button to this page? So that the user can press back if they don't want to create one. We should also clear the errors from the session on the /viewCaps page.

repos/fdbt-site/src/pages/viewCaps.tsx
```diff
@@ -25,20 +35,82 @@ const ViewCap = ({ viewCapErrors = [] }: CapProps): ReactElement => {
...
+                                    >
+                                        Edit
+                                    </a>
+                                </li>
+                            </ul>
+                        </div>
+
+                        <h4 className="govuk-heading-m govuk-!-padding-bottom-4">
+                            {capValidity === 'endOfCalendarDay'
+                                ? 'At the end of a calendar day'
```
RobCattonKPMG, line 83: Could we make this text a little bit smaller, and include the text from the selectCapValidity page here? For example, if they select 'At the end of a calendar day', it would show

At the end of a calendar day (big text)

The cap applies to journeys made before midnight (smaller text, grey colour).

repos/fdbt-site/src/pages/selectCapValidity.tsx
```diff
@@ -58,6 +64,7 @@ export const getFieldset = (errors: ErrorInfo[], endOfFareDay?: string): RadioCo
                     id: 'cap-end-of-service-hint',
                     content: "The cap applies to journeys made during the 'fare day' as defined by your business rules",
```
RobCattonKPMG, line 65: These lines of text could be moved to an object, and exported so it can be used on a different page. For example

export const validityHintText = {
endOfCalendarDay: 'The cap applies to journeys made before midnight',
24hr: 'The cap applies to journeys made within 24hrs of the first tap',
fareDayEnd: "The cap applies to journeys made during the 'fare day' as defined by your business rules"
}

repos/fdbt-site/src/pages/viewCaps.tsx
```diff
@@ -25,20 +35,82 @@ const ViewCap = ({ viewCapErrors = [] }: CapProps): ReactElement => {
...
                     </p>
+
+                    <div>
+                        <h2 className="govuk-heading-l">Cap expiry</h2>
+                        {capValidity ? (
+                            <CapExpiryCard capValidity={capValidity} fareDayEnd={fareDayEnd} />
+                        ) : (
+                            <>
+                                <p className="govuk-body">
+                                    <em>You currently have not set up cap expiry.</em>
```
RobCattonKPMG, line 48:
```suggestion
                                    <em>You currently have no cap expiry saved.</em>
```

repos/fdbt-site/src/pages/selectCapValidity.tsx
NitishAroraKPMG: i don't think the back will make sense, as will be showing them the selected option and they can continue if they want. Let me know, do we still need back button.

repos/fdbt-site/src/data/auroradb.ts
NitishAroraKPMG: it does not throw any error in the console.

repos/fdbt-site/src/pages/selectCapValidity.tsx
```diff
@@ -34,8 +46,9 @@ export const getFieldset = (errors: ErrorInfo[], endOfFareDay?: string): RadioCo
                 label: ' At the end of a calendar day',
```
RobCattonKPMG, line 46: looks to be an extra bit of whitespace at the start of this

repos/fdbt-site/src/pages/selectCapValidity.tsx
```diff
@@ -81,11 +96,12 @@ export const getFieldset = (errors: ErrorInfo[], endOfFareDay?: string): RadioCo
     return CapValidityFieldSet;
 };
 
-const CapValidity = ({ errors = [], fieldset, csrfToken }: CapValidityProps): ReactElement => {
+const CapValidity = ({ capExpirySelected, errors = [], fieldset, csrfToken }: CapValidityProps): ReactElement => {
     return (
         <TwoThirdsLayout title={title} description={description} errors={errors}>
             <CsrfForm action="/api/capValidity" method="post" csrfToken={csrfToken}>
                 <>
+                    {!capExpirySelected ? <BackButton href="/viewCaps" /> : null}
```
RobCattonKPMG, line 104: I think the back button should always appear. User's wont always think that clicking continue would do the same thing.

repos/fdbt-site/src/pages/selectCapValidity.tsx
```diff
@@ -8,18 +8,30 @@ import { getSessionAttribute } from '../utils/sessions';
...
+    '24hr': 'The cap applies to journeys made within 24hrs of the first tap',
+    fareDayEnd: "The cap applies to journeys made during the 'fare day' as defined by your business rules",
+};
+
 interface CapValidityProps {
+    capExpirySelected: string;
     errors: ErrorInfo[];
     fieldset: RadioConditionalInputFieldset;
     csrfToken: string;
 }
```
RobCattonKPMG, line 29: If I select fare day end, and then edit it, the page doesn't show fare day end by default.

RobCattonKPMG: The above isn't happening anymore but if I change the expiry from fare day end to something else, then edit it, it still shows fare day end as the selected value.

NitishAroraKPMG: tested the scenario and working fine now...

repos/fdbt-site/src/data/auroradb.ts
```diff
@@ -1702,6 +1635,52 @@ export const upsertFareDayEnd = async (nocCode: string, fareDayEnd: string): Pro
         throw new Error(`Could not insert fare day end into the fareDayEnd table. ${error}`);
     }
 };
+export const getCapExpiry = async (nocCode: string): Promise<string | undefined> => {
```
RobCattonKPMG, line 1638: Please could you add some commented out examples in this functions of what the stringified version of the capValidity might be? It's hard to tell what we will get out of JSON.parse(string)

repos/fdbt-site/src/pages/selectCapValidity.tsx
```diff
@@ -8,18 +8,29 @@ import { getSessionAttribute } from '../utils/sessions';
...
 import RadioConditionalInput from '../components/RadioConditionalInput';
 import { isCapExpiry } from '../interfaces/typeGuards';
-import { getFareDayEnd } from '../data/auroradb';
+import { getCapExpiry, getFareDayEnd } from '../data/auroradb';
+import BackButton from '../components/BackButton';
 
 const title = 'Cap Validity - Create Fares Data Service';
 const description = 'Cap Validity selection page of the Create Fares Data Service';
 
+export const validityHintText: { [key: string]: string } = {
```
RobCattonKPMG, line 17: `{ [key: string]: string }`

should be 

{ [expiry: string]: string }

repos/fdbt-site/src/pages/selectCapValidity.tsx
```diff
@@ -8,18 +8,29 @@ import { getSessionAttribute } from '../utils/sessions';
 import { getAndValidateNoc, getCsrfToken, getErrorsByIds } from '../utils';
 import RadioConditionalInput from '../components/RadioConditionalInput';
 import { isCapExpiry } from '../interfaces/typeGuards';
-import { getFareDayEnd } from '../data/auroradb';
+import { getCapExpiry, getFareDayEnd } from '../data/auroradb';
+import BackButton from '../components/BackButton';
 
 const title = 'Cap Validity - Create Fares Data Service';
```
RobCattonKPMG, line 14: Can we change all mentions of validity to expiry? That's what we're calling it as it will make more sense to the user than validity.

RobCattonKPMG (changes requested):

RobCattonKPMG (approved):

PR #798, opened 2023-02-14
feat: Add check before deleting cap

NitishAroraKPMG:
## Description

Add the check before deleting the cap

## Testing instructions

Delete the cap after using in a product.


repos/fdbt-site/src/pages/api/deleteCap.ts
```diff
@@ -15,6 +19,26 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
 
+        const products = await getAllProductsByNoc(noc);
+        const matchingJsonLinks = products.map((product) => product.matchingJsonLink);
+        const tickets: TicketWithIds[] = await Promise.all(
+            matchingJsonLinks.map(async (link) => {
+                return await getProductsMatchingJson(link);
+            }),
+        );
+
+        const productsUsingCap = tickets.filter((t) => 'cap' in t && t.cap && t.cap.id === id);
```
RobCattonKPMG, line 30: can we not use 't' please - use ticket instead.

repos/fdbt-site/src/pages/api/deleteCap.ts
```diff
@@ -15,6 +19,26 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+
+        if (productsUsingCap.length > 0) {
+            const cap = (await getCapByNocAndId(noc, id)) as Cap;
```
RobCattonKPMG, line 33: Can we move this line right up to after we check the id is a number, and if it comes back as undefined, throw an error there. Then when we use the cap down here, we dont need to cast it to Cap.

repos/fdbt-site/src/pages/api/deleteCap.ts
```diff
@@ -14,6 +18,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+
+        const products = await getAllProductsByNoc(noc);
+        const matchingJsonLinks = products.map((product) => product.matchingJsonLink);
+        const tickets: TicketWithIds[] = await Promise.all(
+            matchingJsonLinks.map(async (link) => {
+                return await getProductsMatchingJson(link);
+            }),
+        );
+
+        const productsUsingCap = tickets.filter((ticket) => 'cap' in ticket && ticket.cap && ticket.cap.id === id);
```
RobCattonKPMG, line 35:
```suggestion
        const productsUsingCap = tickets.filter((ticket) => ticket.cap && ticket.cap.id === id);
```

repos/fdbt-site/src/pages/api/deleteCap.ts
```diff
@@ -14,6 +18,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
...
+
+        if (productsUsingCap.length > 0) {
+            const { name } = cap.capDetails;
+            const errorMessage = `You cannot delete ${name} because it is being used in ${productsUsingCap.length} ticket(s).`;
+            const errors: ErrorInfo[] = [{ id: 'cap-0', errorMessage }];
+            updateSessionAttribute(req, VIEW_CAP, errors);
+            redirectTo(res, `/viewCaps?cannotDelete=${name}`);
```
RobCattonKPMG, line 42: Why are we redirecting with a query string?

NitishAroraKPMG: this is the way we are doing in other delete api's

repos/fdbt-site/src/pages/api/deleteCap.ts
NitishAroraKPMG: image

RobCattonKPMG: OK no worries - it should work but our types are a bit of a mess.

repos/fdbt-site/src/pages/api/deleteCap.ts
RobCattonKPMG: @NitishAroraKPMG You are correct - but the other delete API's are all also incorrectly doing this. There is nothing on the /viewCaps (or /viewTimeRestrictions etc etc) pages which actually use this query string.

We need to make sure we dont blindly copy things from other parts of the repo, because they aren't always correct.

repos/fdbt-site/src/pages/api/deleteCap.ts
```diff
@@ -14,6 +18,30 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
         }
 
         const noc = getAndValidateNoc(req, res);
+        const cap = await getCapByNocAndId(noc, id);
+
+        if (!cap) {
+            throw new Error('unable to find cap with provided id');
```
alainajaffer, line 24:
```suggestion
            throw new Error('Unable to find cap with provided id');
```

repos/fdbt-site/src/utils/sessions.ts
```diff
@@ -234,6 +235,7 @@ export interface SessionAttributeTypes {
     [VIEW_TIME_RESTRICTION]: ErrorInfo[];
     [VIEW_OPERATOR_GROUP]: ErrorInfo[];
     [VIEW_PRODUCT_GROUP]: ErrorInfo[];
+    [VIEW_CAP]: ErrorInfo[];
```
alainajaffer, line 238:
```suggestion
    [VIEW_CAP_ERRORS]: ErrorInfo[];
```

RobCattonKPMG: Yeah that name makes more sense

NitishAroraKPMG: this is the way the attribute is defined for view pages ...

alainajaffer (approved):

PR #802, opened 2023-02-15
feat: Exporter to swap out caps

alainajaffer:
## Description

The exporter needs to swap out the cap id in the matching JSON for the actual cap from the DB.

## Testing instructions

Code review

### DIFF
--- a/repos/exporter/lib/handler.ts
+++ b/repos/exporter/lib/handler.ts
@@ -9,10 +9,6 @@ import {
     BasePeriodTicket,
     ProductDetails,
     BaseSchemeOperatorTicket,
-    SingleTicket,
-    ReturnTicket,
-    FlatFareTicket,
-    CapSelection,
 } from 'fdbt-types/matchingJsonTypes';
 import {
     getFareDayEnd,
@@ -24,7 +20,7 @@ import {
 } from './database';
 import { ExportLambdaBody } from 'fdbt-types/integrationTypes';
 import 'source-map-support/register';
-import { DbCap, DbTimeRestriction } from 'fdbt-types/dbTypes';
+import { DbTimeRestriction } from 'fdbt-types/dbTypes';
 
 const s3: S3 = new S3(
     process.env.NODE_ENV === 'development'
@@ -94,12 +90,7 @@ export const handler: Handler<ExportLambdaBody> = async ({ paths, noc, exportPre
                 ? await getTimeRestrictionsByIdAndNoc(ticketWithIds.timeRestriction.id, noc)
                 : [];
 
-            const cap =
-                ['single', 'return', 'flatFare'].includes(ticketWithIds.type) &&
-                'cap' in ticketWithIds &&
-                ticketWithIds.cap
-                    ? await getCapByNocAndId(noc, ticketWithIds.cap.id)
-                    : undefined;
+            const cap = !!ticketWithIds.cap ? await getCapByNocAndId(noc, ticketWithIds.cap.id) : undefined;
 
             const fareDayEnd = await getFareDayEnd(noc);
 
--- a/repos/exporter/lib/handler.ts
+++ b/repos/exporter/lib/handler.ts
@@ -9,6 +9,10 @@ import {
     BasePeriodTicket,
     ProductDetails,
     BaseSchemeOperatorTicket,
+    SingleTicket,
+    ReturnTicket,
+    FlatFareTicket,
+    CapSelection,
 } from 'fdbt-types/matchingJsonTypes';
 import {
     getFareDayEnd,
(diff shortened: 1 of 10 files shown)


repos/fdbt-types/dbTypes.ts
```diff
@@ -124,6 +124,13 @@ export interface DbTimeRestriction {
     timeBands: DbTimeBand[];
 }
 
+export interface DbCap {
+    name: string;
+    price: string;
+    durationAmount: string;
+    durationUnits: ExpiryUnit;
```
RobCattonKPMG, line 131: what about capStart? This is sometimes on a cap

repos/fdbt-types/matchingJsonTypes.ts
```diff
@@ -237,6 +238,7 @@ export interface SingleTicket extends BasePointToPointTicket {
     fareZones: FareZone[];
```
RobCattonKPMG, line 238: we need to add to WithBaseIds<T> type, like this

export type WithBaseIds<T> = Omit<
    T,
    | 'passengerType'
    | 'ageRange'
    | 'ageRangeMin'
    | 'ageRangeMax'
    | 'proof'
    | 'proofDocuments'
    | 'timeRestriction'
    | 'fareDayEnd'
> & { passengerType: { id: number }; timeRestriction?: { id: number }; cap?: { id: number } };

repos/fdbt-types/matchingJsonTypes.ts
```diff
@@ -251,6 +253,7 @@ export interface ReturnTicket extends BasePointToPointTicket {
     outboundFareZones: FareZone[];
     returnPeriodValidity?: ReturnPeriodValidity;
     additionalServices?: AdditionalService[];
+    cap?: CapSelection;
```
RobCattonKPMG, line 256: I think we just need to add cap?: CapSelection to BaseTicket, as all other tickets will inherit it.

repos/exporter/lib/handler.ts
```diff
@@ -89,6 +94,13 @@ export const handler: Handler<ExportLambdaBody> = async ({ paths, noc, exportPre
                 ? await getTimeRestrictionsByIdAndNoc(ticketWithIds.timeRestriction.id, noc)
                 : [];
 
+            const cap =
+                ['single', 'return', 'flatFare'].includes(ticketWithIds.type) &&
+                'cap' in ticketWithIds &&
+                ticketWithIds.cap
+                    ? await getCapByNocAndId(noc, ticketWithIds.cap.id)
+                    : undefined;
```
RobCattonKPMG, line 102: We dont need to check the fareType here, because we can trust the site that if there is a cap on the ticket, it should be there. And also in the future, we might change it to be for more than singles, returns and flat fares.

I think we can just do 
```suggestion
            const cap = !!ticketWithIds.cap ? await getCapByNocAndId(noc, ticketWithIds.cap.id) : undefined;
```

repos/fdbt-types/matchingJsonTypes.ts
```diff
@@ -383,3 +386,7 @@ export interface OperatorDetails {
     county: string;
     postcode: string;
 }
+
+export interface CapSelection {
```
RobCattonKPMG, line 390: This will not be needed if we make the above changes

repos/fdbt-types/matchingJsonTypes.ts
```diff
@@ -114,7 +113,7 @@ export type GeoZoneTicket = PeriodGeoZoneTicket | MultiOperatorGeoZoneTicket;
 
 export interface PeriodHybridTicket extends PeriodGeoZoneTicket, PeriodMultipleServicesTicket {}
```
RobCattonKPMG, line 115: We need to do this `export type WithBaseIds<T> = Omit<
    T,
    | 'passengerType'
    | 'ageRange'
    | 'ageRangeMin'
    | 'ageRangeMax'
    | 'proof'
    | 'proofDocuments'
    | 'timeRestriction'
    | 'fareDayEnd'
> & { passengerType: { id: number }; timeRestriction?: { id: number }; cap?: { id: number } };`

repos/fdbt-types/matchingJsonTypes.ts
```diff
@@ -158,6 +157,7 @@ export interface BaseTicket<T extends TicketType = TicketType> {
     ticketPeriod: TicketPeriod;
     groupDefinition?: GroupDefinition;
     carnet?: boolean;
+    caps?: { id: number };
```
RobCattonKPMG, line 160: This is needed because BaseSchemeOperatorTicket extends BaseTicket.

We need to do what is detailed in the above comment, and also edit BaseTicket to have cap?: DbCap

RobCattonKPMG (approved):

PR #810, opened 2023-02-17
fix: Showing the unique services in the service page

NitishAroraKPMG:
## Description

Fix the list of services displayed in service page.

## Testing instructions

View the operator which is having duplicate services in the "View & Manage fares"

### DIFF
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -144,7 +144,7 @@ export const getServicesByNocCodeAndDataSource = async (nocCode: string, source:
 
     try {
         const queryInput = `
-            SELECT id, lineName, lineId, startDate, serviceDescription AS description, origin, destination, serviceCode
+            SELECT id, lineName, lineId, startDate, serviceDescription AS description, origin, destination, serviceCode, endDate
             FROM txcOperatorLine
             WHERE nocCode = ? AND dataSource = ? AND (endDate IS NULL OR CURDATE() <= endDate)
             ORDER BY CAST(lineName AS UNSIGNED) = 0, CAST(lineName AS UNSIGNED), LEFT(lineName, 1), MID(lineName, 2), startDate;
@@ -156,6 +156,7 @@ export const getServicesByNocCodeAndDataSource = async (nocCode: string, source:
             queryResults.map((item) => ({
                 ...item,
                 startDate: convertDateFormat(item.startDate),
+                endDate: item.endDate ? convertDateFormat(item.endDate) : undefined,
             })) || []
         );
     } catch (error) {
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -163,6 +163,38 @@ export const getServicesByNocCodeAndDataSource = async (nocCode: string, source:
     }
 };
 
+export const getFaresServicesByNocCodeAndDataSource = async (
+    nocCode: string,
+    source: string,
+): Promise<MyFaresService[]> => {
+    const nocCodeParameter = replaceInternalNocCode(nocCode);
+    logger.info('', {
+        context: 'data.auroradb',
+        message: 'retrieving services for given noc',
+        noc: nocCode,
+    });
+
+    try {
+        const queryInput = `
+            SELECT id, lineName, lineId, startDate, origin, destination, endDate
+            FROM txcOperatorLine
+            WHERE nocCode = ? AND dataSource = ? AND (endDate IS NULL OR CURDATE() <= endDate)
+            ORDER BY CAST(lineName AS UNSIGNED) = 0, CAST(lineName AS UNSIGNED), LEFT(lineName, 1), MID(lineName, 2), startDate;
+        `;
+
+        const queryResults = await executeQuery<MyFaresService[]>(queryInput, [nocCodeParameter, source]);
+
+        return (
+            queryResults.map((item) => ({
+                ...item,
+                startDate: convertDateFormat(item.startDate),
+            })) || []
+        );
+    } catch (error) {
+        throw new Error(`Could not retrieve services from AuroraDB: ${error.stack}`);
+    }
+};
+
 export const getTndsServicesByNocAndModes = async (nocCode: string, modes: string[]): Promise<ServiceType[]> => {
     const nocCodeParameter = replaceInternalNocCode(nocCode);
     logger.info('', {
--- a/repos/fdbt-site/src/data/auroradb.ts
+++ b/repos/fdbt-site/src/data/auroradb.ts
@@ -164,38 +164,6 @@ export const getServicesByNocCodeAndDataSource = async (nocCode: string, source:
     }
 };
 
-export const getFaresServicesByNocCodeAndDataSource = async (
-    nocCode: string,
-    source: string,
-): Promise<MyFaresService[]> => {
-    const nocCodeParameter = replaceInternalNocCode(nocCode);
-    logger.info('', {
-        context: 'data.auroradb',
-        message: 'retrieving services for given noc',
-        noc: nocCode,
-    });
-
-    try {
-        const queryInput = `
-            SELECT id, lineName, lineId, startDate, origin, destination, endDate
-            FROM txcOperatorLine
-            WHERE nocCode = ? AND dataSource = ? AND (endDate IS NULL OR CURDATE() <= endDate)
-            ORDER BY CAST(lineName AS UNSIGNED) = 0, CAST(lineName AS UNSIGNED), LEFT(lineName, 1), MID(lineName, 2), startDate;
-        `;
-
-        const queryResults = await executeQuery<MyFaresService[]>(queryInput, [nocCodeParameter, source]);
-
-        return (
-            queryResults.map((item) => ({
-                ...item,
-                startDate: convertDateFormat(item.startDate),
-            })) || []
-        );
-    } catch (error) {
-        throw new Error(`Could not retrieve services from AuroraDB: ${error.stack}`);
-    }
-};
-
 export const getTndsServicesByNocAndModes = async (nocCode: string, modes: string[]): Promise<ServiceType[]> => {
     const nocCodeParameter = replaceInternalNocCode(nocCode);
     logger.info('', {
--- a/repos/fdbt-site/src/pages/products/services.tsx
+++ b/repos/fdbt-site/src/pages/products/services.tsx
@@ -7,7 +7,7 @@ import {
 } from '../../interfaces/index';
 import { BaseLayout } from '../../layout/Layout';
 import { getPointToPointProducts, getBodsOrTndsServicesByNoc } from '../../data/auroradb';
-import { getAndValidateNoc, getMyUniqueServices } from '../../utils';
+import { getAndValidateNoc, removeDuplicateServices } from '../../utils';
 import moment from 'moment';
 import { MyFaresProduct } from '../../interfaces/dbTypes';
 import { getSessionAttribute } from '../../utils/sessions';
@@ -188,8 +188,12 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
     const noc = getAndValidateNoc(ctx);
     const dataSource = !!getSessionAttribute(ctx.req, MULTI_MODAL_ATTRIBUTE) ? 'tnds' : 'bods';
     const services: MyFaresService[] = await getBodsOrTndsServicesByNoc(noc, dataSource);
-    const unqiueServiceIds = getMyUniqueServices(services);
-    const servicesWithNoDuplicates = services.filter((service) => unqiueServiceIds.includes(service.id));
+    const servicesWithNoDuplicates = removeDuplicateServices<MyFaresService>(
+        services,
+        'lineId',
+        'startDate',
+        'endDate',
+    );
 
     const products = await getPointToPointProducts(noc);
     const servicesAndProducts = matchProductsToServices(servicesWithNoDuplicates, products);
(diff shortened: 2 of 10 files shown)


repos/fdbt-site/src/pages/products/services.tsx
```diff
@@ -187,7 +187,18 @@ export const matchProductsToServices = (
...
     const noc = getAndValidateNoc(ctx);
     const dataSource = !!getSessionAttribute(ctx.req, MULTI_MODAL_ATTRIBUTE) ? 'tnds' : 'bods';
-    const services: MyFaresService[] = await getBodsOrTndsServicesByNoc(noc, dataSource);
+    const allServices: MyFaresService[] = await getBodsOrTndsServicesByNoc(noc, dataSource);
+    const uniqueServiceList: string[] = [];
+    const services: MyFaresService[] = [];
+
+    allServices.forEach((service) => {
+        const service_name = `${service.lineId}#${service.startDate}#${service.endDate}`;
+        if (!uniqueServiceList.includes(service_name)) {
```
RobCattonKPMG, line 196: service_name is a python naming standard, we use camelCase in JS/TS.

We need a function which takes an array of services as an argument, and returns an array of services, but removes any duplicates (any services that share the same lineId, startDate and endDate).

Then we need to write unit tests for this function, with different scenarios.

RobCattonKPMG: Then the usage on this page will be 

` const services: MyFaresService[] = await getBodsOrTndsServicesByNoc(noc, dataSource);`
`const servicesWithNoDuplicates = removeDuplicateServices(services);`

repos/fdbt-site/src/data/auroradb.ts
```diff
@@ -163,6 +163,38 @@ export const getServicesByNocCodeAndDataSource = async (nocCode: string, source:
     }
 };
 
+export const getFaresServicesByNocCodeAndDataSource = async (
```
RobCattonKPMG, line 166: Instead of copy and pasting an existing DB query and adding 1 additional piece of information to select, cant we just edit the current DB query (getServicesByNocCodeAndDataSource) and edit the interface ServiceType to include endDate?

RobCattonKPMG: Also we presumably need to use the 'convertDateFormat' function on the end date, if its there, like we do with the start date in the query results.

repos/fdbt-site/src/utils/index.ts
```diff
@@ -258,6 +259,25 @@ export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName:
...
-export const fareTypeIsAllowedToAddACap = (fareType: string): boolean => {
+export const fareTypeIsAllowedToAddACap = (fareType: string | undefined): boolean => {
     return !!fareType && ['single', 'return', 'flatFare'].includes(fareType);
 };
+
+export const getUniqueServices = (services: MyFaresService[]): MyFaresService[] => {
+    const uniqueServiceList: string[] = [];
+    const uniqueServices: MyFaresService[] = [];
+
+    if (!!services && services.length > 0) {
```
RobCattonKPMG, line 270: If the function's arguments says it accepts an array of services as an input, we dont need !!services - we know it will be defines. We also dont need to check the length, because if we use a map function, it will return an empty array if it begins with an empty array.

repos/fdbt-site/src/pages/service.tsx
```diff
@@ -24,7 +24,7 @@ const errorId = 'service';
 interface ServiceProps {
     operator: string;
     passengerType: string;
-    services: ServiceType[];
```
RobCattonKPMG, line 27: I dont think we need to change this type to be MyFaresService - just need to add endDate to ServiceType

repos/fdbt-site/src/utils/index.ts
```diff
@@ -258,6 +259,25 @@ export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName:
     return exportNamePart === exportName;
 };
 
-export const fareTypeIsAllowedToAddACap = (fareType: string): boolean => {
+export const fareTypeIsAllowedToAddACap = (fareType: string | undefined): boolean => {
```
RobCattonKPMG, line 262: Why have we changed this to allow undefined to be passed to it?

repos/fdbt-site/tests/pages/service.test.tsx
```diff
@@ -231,32 +230,31 @@ describe('pages', () => {
...
                         {
-                            id: 12,
+                            id: '12',
                             lineName: 'X1',
                             lineId: '3h3vb32ik',
                             startDate: '06/02/2020',
-                            description: 'this bus service is X1',
                             origin: 'Edinburgh',
-                            serviceCode: 'NW_05_BLAC_X1_1',
+                            destination: 'Leeds',
```
RobCattonKPMG, line 247: The fact that all these changed should be setting off alarm bells that we're doing something wrong. We no longer have descripton in the data, and we dont have the serviceCode either. This could (and probably does) break things further down the line.

repos/fdbt-site/tests/utils/index.test.ts
```diff
@@ -249,4 +252,66 @@ describe('index', () => {
...
+                origin: 'Vega de Pas',
+                destination: 'Guiddam',
+                startDate: '05/04/2020',
+                endDate: '25/04/2020',
+            },
+        ];
+
+        it('returns unique services if two services having same lineName, startDate and endDate ', () => {
+            const result = getUniqueServices(services);
+            expect(services.length).toBeLessThanOrEqual(4);
```
RobCattonKPMG, line 313: We should be checking they return the services we expect, not just checking the length.

RobCattonKPMG: And we also need a test to make sure it doesnt change anything if there are no duplicates.

repos/fdbt-site/src/data/auroradb.ts
NitishAroraKPMG: getServicesByNocCodeAndDataSource -> gives the list of active services (whose endDate is either not present and if present should be greater than current date)

getBodsOrTndsServicesByNoc -> gets all the services.

repos/fdbt-site/src/utils/index.ts
NitishAroraKPMG: ok

repos/fdbt-site/src/data/auroradb.ts
RobCattonKPMG: I understand what the DB queries do, but I dont understand why that means why we cant do what i said above?

RobCattonKPMG: @NitishAroraKPMG question remains - why do we need to add a new DB query? Why can we not just edit the ones used before?

repos/fdbt-site/src/utils/index.ts
```diff
@@ -261,3 +263,33 @@ export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName:
...
+        if (!uniqueServiceList.includes(serviceName)) {
+            uniqueServices.push(service);
+            uniqueServiceList.push(serviceName);
+        }
+    });
+
+    return uniqueServices;
+};
+
+export const getUniqueFaresServices = (services: MyFaresService[]): MyFaresService[] => {
```
RobCattonKPMG, line 282:
```suggestion
export const getUniqueMyFaresServices = (services: MyFaresService[]): MyFaresService[] => {
```

repos/fdbt-site/src/utils/index.ts
```diff
@@ -261,3 +263,33 @@ export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName:
 export const fareTypeIsAllowedToAddACap = (fareType: string): boolean => {
     return !!fareType && ['single', 'return', 'flatFare'].includes(fareType);
 };
+
+export const getUniqueServices = (services: ServiceType[]): ServiceType[] => {
```
RobCattonKPMG, line 267: Can we not just have one function which does this, but it just takes this as an argument:

`{ lineId: string, startDate: string, endDate: string | undefined } []`

And then you can just pass the MyFaresService[] and the ServiceType[] to the function, and it wont care that they have additional information?

NitishAroraKPMG: what about the return from the function

repos/fdbt-site/src/data/auroradb.ts
NitishAroraKPMG: will be deleting the function in next push as it is not referenced anywhere

repos/fdbt-site/src/pages/products/services.tsx
```diff
@@ -188,8 +188,11 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
     const noc = getAndValidateNoc(ctx);
     const dataSource = !!getSessionAttribute(ctx.req, MULTI_MODAL_ATTRIBUTE) ? 'tnds' : 'bods';
     const services: MyFaresService[] = await getBodsOrTndsServicesByNoc(noc, dataSource);
+    const unqiueServiceIds = getMyUniqueServices(services);
```
RobCattonKPMG, line 191:
```suggestion
    const uniqueServiceIds = getUniqueServiceIds(services);
```

repos/fdbt-site/src/utils/index.ts
```diff
@@ -261,3 +263,50 @@ export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName:
 export const fareTypeIsAllowedToAddACap = (fareType: string): boolean => {
     return !!fareType && ['single', 'return', 'flatFare'].includes(fareType);
 };
+
+export const getMyUniqueServices = (
+    services: { id?: string | number; lineId: string; startDate: string; endDate?: string }[],
```
RobCattonKPMG, line 268: If we're using id's to store which ones are unique, why is id nullable on the input?

repos/fdbt-site/src/utils/index.ts
```diff
@@ -261,3 +263,50 @@ export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName:
...
+): string[] => {
+    const uniqueServiceList: string[] = [];
+    const uniqueServices: string[] = [];
+
+    services.forEach((service) => {
+        const serviceName = `${service.lineId}#${service.startDate}#${service.endDate ? service.endDate : ''}`;
+        if (!uniqueServiceList.includes(serviceName)) {
+            uniqueServices.push(String(service.id));
```
RobCattonKPMG, line 276: are id's not strings?

repos/fdbt-site/src/pages/service.tsx
```diff
@@ -108,9 +108,13 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
 
     const dataSourceAttribute = getRequiredSessionAttribute(ctx.req, TXC_SOURCE_ATTRIBUTE);
 
-    const services = await getServicesByNocCodeAndDataSource(nocCode, dataSourceAttribute.source);
+    const services: ServiceType[] = await getServicesByNocCodeAndDataSource(nocCode, dataSourceAttribute.source);
+    const unqiueServiceIds = getMyUniqueServices(services);
+    const servicesWithNoDuplicates = services.filter(
```
RobCattonKPMG, line 113: Is there a reason why we cant move this filter into the function which gets unique services? Why do we have to do it here?

repos/fdbt-site/src/utils/index.ts
```diff
@@ -261,3 +263,50 @@ export const objectKeyMatchesExportNameExactly = (objectKey: string, exportName:
...
+    });
+
+    return uniqueServices;
+};
+
+export const getUniqueServices = (services: ServiceType[]): ServiceType[] => {
+    const uniqueServiceList: string[] = [];
+    const uniqueServices: ServiceType[] = [];
+
+    services.forEach((service) => {
```
RobCattonKPMG, line 288: I dont think it makes sense for us to have the same code 3 times over, slightly altered.

repos/fdbt-site/src/utils/index.ts
RobCattonKPMG: @NitishAroraKPMG needs a response please

repos/fdbt-site/tests/utils/index.test.ts
```diff
@@ -249,4 +251,136 @@ describe('index', () => {
...
+                    lineId: '3h3vb32ik',
+                    startDate: '06/02/2020',
+                    description: 'this bus service is X1',
+                    origin: 'Edinburgh',
+                    serviceCode: 'NW_05_BLAC_X1_1',
+                },
+            ]);
+        });
+
+        const mockFaresServices = [
```
RobCattonKPMG, line 331: Can we move this to the top of the describe block please

RobCattonKPMG: And can we also add the type of MyFaresService[] to this variable

repos/fdbt-site/tests/utils/index.test.ts
```diff
@@ -249,4 +251,136 @@ describe('index', () => {
...
+        });
+
+        it.each(unCapFareTypes)('returns false if the fare type is other than %s', (fareType: string) => {
+            const result = fareTypeIsAllowedToAddACap(fareType);
+            expect(result).toBeFalsy();
+        });
+    });
+
+    describe('removeDuplicateServices', () => {
+        const mockServices = [
```
RobCattonKPMG, line 271: ServiceType type needs adding to this variable

repos/fdbt-site/tests/utils/index.test.ts
```diff
@@ -249,4 +252,136 @@ describe('index', () => {
...
+            expect(result).toBeTruthy();
+        });
+
+        it.each(unCapFareTypes)('returns false if the fare type is other than %s', (fareType: string) => {
+            const result = fareTypeIsAllowedToAddACap(fareType);
+            expect(result).toBeFalsy();
+        });
+    });
+
+    describe.only('removeDuplicateServices', () => {
```
RobCattonKPMG, line 271:
```suggestion
    describe('removeDuplicateServices', () => {
```

RobCattonKPMG (approved):

PR #818, opened 2023-02-23
fix: Not showing the export button after the select export

NitishAroraKPMG:
## Description

Users should be unable to see the export button just after selecting the products to export

## Testing instructions

Click on few products to export and just after that, click on the Export all products. Now, the export all products button will not appear just after the select product export completed.

### DIFF
--- a/repos/fdbt-site/src/pages/api/getExportProgress.ts
+++ b/repos/fdbt-site/src/pages/api/getExportProgress.ts
@@ -1,5 +1,5 @@
 import { NextApiResponse } from 'next';
-import { dateIsOverThirtyMinutesAgo, getAndValidateNoc, redirectToError } from '../../utils/apiUtils';
+import { dateIsOverThirtyMinutesAgo, exportHasStarted, getAndValidateNoc, redirectToError } from '../../utils/apiUtils';
 import { NextApiRequestWithSession } from '../../interfaces';
 import {
     checkIfMetaDataExists,
@@ -30,9 +30,8 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
 
         const exportNames = await getS3Exports(noc);
         const selectExportAttribute = getSessionAttribute(req, SELECT_EXPORTS_ATTRIBUTE);
-        const currTime = new Date().getTime() / 1000;
 
-        if (!!selectExportAttribute && currTime - Number(selectExportAttribute.exportStarted) > 5) {
+        if (!!selectExportAttribute && exportHasStarted(selectExportAttribute.exportStarted)) {
             updateSessionAttribute(req, SELECT_EXPORTS_ATTRIBUTE, undefined);
         }
 
--- a/repos/fdbt-site/src/pages/api/getExportProgress.ts
+++ b/repos/fdbt-site/src/pages/api/getExportProgress.ts
@@ -12,6 +12,8 @@ import {
 import { MATCHING_DATA_BUCKET_NAME, NETEX_BUCKET_NAME } from '../../constants';
 import logger from '../../utils/logger';
 import { difference } from 'lodash';
+import { getSessionAttribute, updateSessionAttribute } from '../../utils/sessions';
+import { SELECT_EXPORTS_ATTRIBUTE } from '../../constants/attributes';
 
 export interface Export {
     name: string;
@@ -27,6 +29,12 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
         const noc = getAndValidateNoc(req, res);
 
         const exportNames = await getS3Exports(noc);
+        const selectExportAttribute = getSessionAttribute(req, SELECT_EXPORTS_ATTRIBUTE);
+        const currTime = new Date().getTime() / 1000;
+
+        if (!!selectExportAttribute && currTime - Number(selectExportAttribute.exportStarted) > 5) {
+            updateSessionAttribute(req, SELECT_EXPORTS_ATTRIBUTE, undefined);
+        }
 
         const exports: Export[] = await Promise.all(
             exportNames.map(async (name) => {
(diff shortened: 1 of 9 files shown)


repos/fdbt-site/src/pages/products/exports.tsx
```diff
@@ -52,21 +53,25 @@ const getTag = (exportDetails: Export): ReactElement => {
...
-const Exports = ({ csrf, operatorHasProducts }: GlobalSettingsProps): ReactElement => {
+const Exports = ({ csrf, exportStarted, operatorHasProducts }: GlobalSettingsProps): ReactElement => {
     const [showExportPopup, setShowExportPopup] = useState(false);
     const [showFailedFilesPopup, setShowFailedFilesPopup] = useState(false);
     const [buttonClicked, setButtonClicked] = useState(false);
+    const [isExportStarted, setIsExportStarted] = useState(exportStarted);
 
     const { data } = useSWR('/api/getExportProgress', fetcher, { refreshInterval: 1500 });
 
+    setInterval(() => setIsExportStarted(false), 1000 * 1500);
```
RobCattonKPMG, line 64: What is the intention of this line of code?

repos/fdbt-site/src/pages/products/exports.tsx
```diff
@@ -52,21 +53,25 @@ const getTag = (exportDetails: Export): ReactElement => {
     );
 };
 
-const Exports = ({ csrf, operatorHasProducts }: GlobalSettingsProps): ReactElement => {
+const Exports = ({ csrf, exportStarted, operatorHasProducts }: GlobalSettingsProps): ReactElement => {
     const [showExportPopup, setShowExportPopup] = useState(false);
     const [showFailedFilesPopup, setShowFailedFilesPopup] = useState(false);
     const [buttonClicked, setButtonClicked] = useState(false);
+    const [isExportStarted, setIsExportStarted] = useState(exportStarted);
```
RobCattonKPMG, line 60: instead of isExportStarted, lets call it 'hasExportStarted' and 'setHasExportStarted'

repos/fdbt-site/src/pages/products/exports.tsx
NitishAroraKPMG: to set the state of flag to false after 1500 secs

repos/fdbt-site/src/pages/api/getExportProgress.ts
```diff
@@ -27,6 +29,12 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
         const noc = getAndValidateNoc(req, res);
 
         const exportNames = await getS3Exports(noc);
+        const selectExportAttribute = getSessionAttribute(req, SELECT_EXPORTS_ATTRIBUTE);
+        const currTime = new Date().getTime() / 1000;
```
RobCattonKPMG, line 33: if I run new Date().getTime(), I get 1677164819135 . If I divide that by 1000, i get 1677164825.979 , which will always be more than 5.

RobCattonKPMG: Also, why is this code in the getExportProgress API? We need to be doing this inside exports.tsx

repos/fdbt-site/src/pages/api/selectExports.ts
```diff
@@ -40,7 +42,8 @@ export default async (req: NextApiRequestWithSession, res: NextApiResponse): Pro
     const links = products.map((product) => product.matchingJsonLink);
 
     await triggerExport({ noc, paths: links, exportPrefix: exportName });
-
-    redirectTo(res, '/products/exports');
+    const currTime = new Date().getTime() / 1000;
```
RobCattonKPMG, line 45: I dont understand how this works - new Date().getTime() returns 1677164983380 , why do we divide that by 1000?

NitishAroraKPMG: to get the time in seconds

repos/fdbt-site/src/pages/api/getExportProgress.ts
RobCattonKPMG: @NitishAroraKPMG ok but the comment above is asking why this is in the getExportProgress API?

repos/fdbt-site/src/pages/api/selectExports.ts
RobCattonKPMG: Makes sense - but it shouldnt be in this API.

repos/fdbt-site/src/utils/sessions.ts
```diff
@@ -251,6 +252,7 @@ export interface SessionAttributeTypes {
     [CAPS_DEFINITION_ATTRIBUTE]: CapSelection | { errors: ErrorInfo[] };
     [SERVICE_LIST_EXEMPTION_ATTRIBUTE]: ServiceListAttribute | { errors: ErrorInfo[] };
     [STOPS_EXEMPTION_ATTRIBUTE]: ExemptedStopsAttribute | { errors: ErrorInfo[] };
+    [SELECT_EXPORTS_ATTRIBUTE]: { exportStarted: string };
```
RobCattonKPMG, line 255: Seems more appropriate for this to be a number rather than a string

repos/fdbt-site/src/pages/products/exports.tsx
```diff
@@ -222,10 +226,19 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
     const noc = getAndValidateNoc(ctx);
 
     const operatorHasProducts = (await getAllProductsByNoc(noc)).length > 0;
+    const selectExportAttribute = getSessionAttribute(ctx.req, SELECT_EXPORTS_ATTRIBUTE);
+    const currTime = new Date().getTime() / 1000;
+
+    const exportStarted = !!selectExportAttribute && currTime - Number(selectExportAttribute.exportStarted) > 5;
+
```
RobCattonKPMG, line 233: I think we should move this logic into a function called exportHasStarted, and then we can check it via a unit test. And it means we can use the function in multiple places.

NitishAroraKPMG: was also thinking the same and it can be called at multiple places

RobCattonKPMG (approved):

PR #821, opened 2023-02-27
fix: Showing the services when end type is null

NitishAroraKPMG:
## Description

Display the services for operator having source of data as tnds and the end date as null

## Testing instructions

Find the ferry operator with the service whose end date is not populated in db.

### DIFF
--- a/repos/fdbt-site/src/interfaces/index.ts
+++ b/repos/fdbt-site/src/interfaces/index.ts
@@ -597,8 +597,9 @@ export interface ServiceType {
     serviceCode: string;
     dataSource?: string;
     mode?: string;
-    endDate?: string;
+    endDate?: string | null;
 }
+
 export interface ServiceCount {
     serviceCount: number;
 }
--- a/repos/fdbt-site/src/interfaces/index.ts
+++ b/repos/fdbt-site/src/interfaces/index.ts
@@ -597,7 +597,7 @@ export interface ServiceType {
     serviceCode: string;
     dataSource?: string;
     mode?: string;
-    endDate?: string | null;
+    endDate: string | null;
 }
 
 export interface ServiceCount {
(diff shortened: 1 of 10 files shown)


repos/fdbt-site/src/interfaces/index.ts
```diff
@@ -597,8 +597,9 @@ export interface ServiceType {
     serviceCode: string;
     dataSource?: string;
     mode?: string;
-    endDate?: string;
+    endDate?: string | null;
```
RobCattonKPMG, line 600: Do we need the '?' on the endDate now? Will it ever be undefined?

repos/fdbt-site/tests/pages/api/fareType.test.ts
```diff
@@ -28,6 +29,7 @@ const services = [
         origin: 'Bolton',
         destination: 'Wigan',
         dataSource: 'bods',
```
RobCattonKPMG, line 31: This service has dataSource bods, which means endDate wont be null

repos/fdbt-site/tests/pages/home.test.tsx
```diff
@@ -93,6 +96,7 @@ describe('pages', () => {
                     destination: 'Leeds',
                     dataSource: 'bods',
                     mode: 'ferry',
+                    endDate: null,
```
RobCattonKPMG, line 99: This service has dataSource bods, which means endDate wont be null

repos/fdbt-site/tests/pages/home.test.tsx
```diff
@@ -128,6 +132,7 @@ describe('pages', () => {
...
```
RobCattonKPMG, line 135: This service has dataSource bods, which means endDate wont be null

RobCattonKPMG (approved):

PR #822, opened 2023-02-27
fix: moving xsl transform to validator

RobCattonKPMG:
## Description

Previously, an xsl transformation to the xml was happening in the netex converter. Due to Ubuntu 18.02 being deprecated soon, we are moving to 20.04 and the libxml library we were using to run the xslt would not work on 18.02. As such, we have moved the xslt into the netex validator instead.

## Testing instructions

Run an export.


repos/fdbt-netex-output/package.json
```diff
@@ -23,7 +23,7 @@
         "aws-param-store": "^3.2.0",
         "aws-sdk": "^2.938.0",
         "fdbt-types": "file:../fdbt-types",
-        "libxslt": "^0.10.0",
+        "libxslt": "^0.10.1",
```
lj-kpmg, line 26: Can this be moved to devDependencies now?

repos/fdbt-netex-output/src/netex-convertor/handler.ts
```diff
@@ -38,17 +37,6 @@ export const xsl = `
     </xsl:stylesheet>
```
lj-kpmg, line 37: This xsl can be deleted now

repos/fdbt-netex-output/src/netex-validator/main.py
```diff
@@ -99,13 +130,28 @@ def lambda_handler(event, context):
 
         try:
             logger.info('Starting NeTEx validation...')
-            validate_netex(download_path)
 
-            logger.info('NeTEx valid, uploading to S3...')
-            s3_client.upload_file(download_path, os.getenv('VALIDATED_NETEX_BUCKET'), key)
+            netex_file = open(download_path, 'r+', encoding="ISO-8859-1")
```
lj-kpmg, line 134: This can just be r since we're only reading

repos/fdbt-netex-output/src/netex-validator/main.py
```diff
@@ -99,13 +130,28 @@ def lambda_handler(event, context):
...
+            netex = netex_file.read()
+            netex_file.close()
+            transformed_netex = transform_netex_with_xsl(netex)
+            validate_netex(transformed_netex)
+
+            logger.info('NeTEx valid, saving transformed netex...')
+
+            transformed_netex_file = open(
+                download_path, 'w+', encoding="ISO-8859-1")
```
lj-kpmg, line 143: This can just be w since we're only writing

repos/fdbt-netex-output/src/netex-convertor/handler.ts
RobCattonKPMG: Moved it to generateAll file

repos/fdbt-netex-output/package.json
RobCattonKPMG: Yep, moved.

repos/fdbt-netex-output/src/netex-validator/main.py
RobCattonKPMG: Will change

repos/fdbt-netex-output/src/netex-validator/main.py
RobCattonKPMG: will change

lj-kpmg (approved):

PR #827, opened 2023-03-07
fix: Fixing the bug for the change in the redirection of point to point

NitishAroraKPMG:
## Description

The "change" in ticket confirmation is redirecting the user to multipleProducts instead of pointToPointPeriodProduct

## Testing instructions

Create a point to point period/school ticket and click on Change in the ticket confirmation page.

### DIFF
--- a/repos/fdbt-site/tests/pages/ticketConfirmation.test.tsx
+++ b/repos/fdbt-site/tests/pages/ticketConfirmation.test.tsx
@@ -1,6 +1,6 @@
 import { shallow } from 'enzyme';
 import * as React from 'react';
-import { ExpiryUnit } from '../../src/interfaces/matchingJsonTypes';
+import { ExpiryUnit, PointToPointPeriodProduct, Product } from '../../src/interfaces/matchingJsonTypes';
 import {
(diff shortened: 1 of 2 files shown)


repos/fdbt-site/tests/pages/ticketConfirmation.test.tsx
```diff
@@ -498,7 +499,7 @@ describe('pages', () => {
                 });
                 const confirmationElements = buildSchoolTicketConfirmationElements(ctx);
                 expect(confirmationElements).toContainEqual(confirmationElementStructure);
-                expect(confirmationElements).toHaveLength(2);
+                expect(confirmationElements).toHaveLength(13);
```
RobCattonKPMG, line 502: I assume this number has increased by 11 because its generating confirmation elements for a point to point period. This test needs renaming to show that, and we will need a new test for a non-period point to point academic ticket.

NitishAroraKPMG: The number is incremented as the function calling is changed.

RobCattonKPMG (approved):

PR #829, opened 2023-03-08
Getting the export progress in page load

NitishAroraKPMG:
## Description

Avoiding the users to click 2 exports

## Testing instructions

Select products to export and then Click on "Export All" for the user

### DIFF
--- a/repos/fdbt-site/src/pages/products/exports.tsx
+++ b/repos/fdbt-site/src/pages/products/exports.tsx
@@ -11,6 +11,8 @@ import InfoPopup from '../../components/InfoPopup';
 import { getSessionAttribute, updateSessionAttribute } from '../../utils/sessions';
 import { SELECT_EXPORTS_ATTRIBUTE } from '../../constants/attributes';
 import { exportHasStarted } from '../../utils/apiUtils';
+import { checkIfMetaDataExists, getExportMetaData, getS3Exports, getS3FolderCount } from '../../data/s3';
+import { MATCHING_DATA_BUCKET_NAME, NETEX_BUCKET_NAME } from '../../constants';
 
 const title = 'Exports';
 const description = 'Export your products into NeTEx.';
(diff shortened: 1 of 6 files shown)


repos/fdbt-site/src/pages/products/exports.tsx
```diff
@@ -234,11 +236,37 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
     if (exportStarted) {
         updateSessionAttribute(ctx.req, SELECT_EXPORTS_ATTRIBUTE, undefined);
     }
+    const exportNames = await getS3Exports(noc);
```
RobCattonKPMG, line 239: If we are no longer using the code above this (the SELECT_EXPORTS_ATTRIBUTE etc) then we need to remove it, from here and also from the API where it is set.

repos/fdbt-site/src/pages/products/exports.tsx
```diff
@@ -234,11 +236,37 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
...
+            const matchingDataCount = await getS3FolderCount(MATCHING_DATA_BUCKET_NAME, prefix);
+            const netexCount = await getS3FolderCount(NETEX_BUCKET_NAME, prefix);
+
+            let numberOfFilesExpected = matchingDataCount;
+
+            let metadata = undefined;
+            const metaDataExists = await checkIfMetaDataExists(`${noc}/exports/${name}.json`);
+
+            if (metaDataExists) {
+                metadata = await getExportMetaData(`${noc}/exports/${name}.json`);
```
RobCattonKPMG, line 254: The problem with this approach, is that we will figure out if an export is in progress, and pass that through to the page as a prop (true or false).

And then that prop will never be updated in the page, because its not using state. 

We need a line of code above like this:

const [exportInProgress, setExportInProgress] = useState(exportInProgress);

So when the page loads, it will have the default value from the page props, and then we will need some code which updates this value.

PR #844, opened 2023-04-03
feature: missing stops page

RobCattonKPMG:
## Description

New page to explain to the user why their service cannot be used to create Netex - because there is a missing stop in the Naptan database that needs to be added.

## Testing instructions

Attempt to create a product with a service that has a stop in the service data which is not in the Naptan dataset.

### DIFF
--- a/repos/fdbt-site/src/pages/editFareStageMatching.tsx
+++ b/repos/fdbt-site/src/pages/editFareStageMatching.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement, useState } from 'react';
 import BackButton from '../components/BackButton';
 import { renderResetAndAutoPopulateButtons, StopItem } from '../components/MatchingBase';
-import { batchGetStopsByAtcoCode, getServiceByIdAndDataSource } from '../data/auroradb';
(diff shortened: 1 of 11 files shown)


repos/fdbt-site/src/pages/editFareStageMatching.tsx
```diff
@@ -381,10 +383,17 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
...
 
     // filling out stop information from DB
-    const naptanInfo = await batchGetStopsByAtcoCode(
+    const dbResults = await batchGetStopsByAtcoCodeWithErrorCheck(
         masterStopList.filter((stop, index, self) => self.indexOf(stop) === index),
     );
 
+    if (!dbResults.successful && ctx.res) {
+        updateSessionAttribute(ctx.req, MISSING_STOPS_ATTRIBUTE, dbResults.missingStops);
+        redirectTo(ctx.res, '/missingStops');
```
lj-kpmg, line 392: You maybe want a return after this

RobCattonKPMG: added

lj-kpmg (approved):

PR #854, opened 2023-07-03
fix: update lambdas to prevent table renamer from running when error

alainajaffer:
## Description

Update RDS where a failure in the retrieval function for either BODS or TNDS does not delete existing tables 

## Testing instructions

None

### DIFF
--- a/repos/exporter/lib/referenceData/tableRenameHandler.ts
+++ b/repos/exporter/lib/referenceData/tableRenameHandler.ts
@@ -5,33 +5,40 @@ import { getConnection, checkReferenceDataImportHasCompleted, deleteAndRenameTab
 import { getSsmValue, putParameter } from '../ssm';
 
 export const handler: Handler<ZipperLambdaBody> = async () => {
-    console.log('triggered reference data tables rename lambda... ');
-
-    const connection = await getConnection();
-
-    let disableRenamer: string | undefined = 'true';
     try {
-        disableRenamer = await getSsmValue('/scheduled/disable-table-renamer');
-    } catch (error) {
-        if (error instanceof Error) {
-            throw new Error(`Failed to get parameter from ssm: ${error.stack || ''}`);
+        console.log('triggered reference data tables rename lambda... ');
+
+        const connection = await getConnection();
+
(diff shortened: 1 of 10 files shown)


repos/fdbt-reference-data-service/src/uploaders/txc_uploader/main.py
```diff
@@ -39,6 +39,12 @@ def lambda_handler(event, context):
             s3_client, cloudwatch_client, bucket, key, file_path, db_connection, logger
         )
     except Exception as e:
+        ssm_client.put_parameter(
```
alvak1: As this lambda is invoked for every file uploaded to S3 - a system wide issue will flood SSM api. The `checkReferenceDataImportHasCompleted` function in table renamer will be able to identify any system wide issues impacting this lambda

repos/fdbt-reference-data-service/src/retrievers/txc_unzipper/main.py
```diff
@@ -54,6 +55,12 @@ def lambda_handler(event, context):
             Namespace='FDBT/Reference-Data-Retrievers'
         )
     except Exception as e:
+        ssm_client.put_parameter(
```
alvak1: As this lambda is invoked for every file uploaded to S3 - a system wide issue will flood SSM api.

repos/exporter/lib/referenceData/tableRenameHandler.ts
```diff
@@ -2,17 +2,36 @@ import { Handler } from 'aws-lambda';
...
+        await checkReferenceDataImportHasCompleted('naptanStop', connection);
+
+        await deleteAndRenameTables(connection);
+
+        await connection.end();
+    } else {
+        throw new Error(
+            'The SSM Parameter used to check for errors in the scheduled job has returned TRUE indicating an issue',
+        );
+    }
```
alvak1, line 35: Can we enclose this in a try/catch block so that in case of an exception the ssm param is still set to false. Preferably in a finally block. This is cause the value is reset only as this step and if there is uncaught exception it will impact the next day run.

PR #899, opened 2024-02-07
fix: Stops / Lines not correct

alainajaffer:
## Description
Unexpected stops for a line are appearing in the stops lists.

Expected:

NVTR 88 stops list is correct

Actual:

NVTR 88 is showing the following stops which are not part of it’s route:

•	Maidstone, opp Bramble Close (on Queen's Road)                             kntagajm
•	Maidstone, adj Malling Terrace (on Queen's Road)                             kntagajt
•	Maidstone, opp Saxon Mews (on Queen's Road)                                 kntagama
•	Maidstone, adj Warden Close (on Queen's Road)                               kntagamd
•	Maidstone, adj Court Drive (on Queen's Road)                                    kntagamj
•	Maidstone, adj Queen's Avenue (on Queen's Road)                           kntagamt
•	Maidstone, Stop A Somerfield Hospital (on London Road)               kntgdpwa

Elab:

Stop lists for a line should be constructed by following Line > VehicleJourney > JourneyPatterns > StopPoints

Note: this approach only applies to BODS derived TransXChange data. The logic for creating stop lists from TNDS derived data should remain the same


repos/fdbt-reference-data-service/src/uploaders/txc_uploader/txc_processor.py
```diff
@@ -177,45 +186,107 @@ def collect_journey_patterns(data: dict, service: dict):
...
+        ):
+            continue
+
         journey_pattern_id = insert_into_txc_journey_pattern_table(
-            cursor, operator_service_id, journey_pattern["journey_pattern_info"]
+            cursor, operator_service_id, journey_pattern_info
         )
 
         links = []
+        stop_codes = set()
```
ljones92, line 263: You should be able to get rid of this I think since it's only used in disruptions for getting the admin area codes

repos/fdbt-reference-data-service/src/uploaders/txc_uploader/txc_processor.py
```diff
@@ -358,6 +436,72 @@ def all_journey_pattern_sections_are_not_empty(data: dict, service: dict) -> boo
...
+        and (
+            ("LineRef" in journey and journey["LineRef"] == line_id)
+            or (
+                "VehicleJourneyCode" in journey
+                and journey["VehicleJourneyCode"] in vehicle_journey_refs
+            )
+        )
+    ]
+
+    vehicle_journeys_data = []
```
ljones92, line 486: `journey_pattern_count` won't be needed here since that's used for tracks data so this line onwards could be simplified to:

```python
return [
    collect_vehicle_journey(vehicle_journey)
    for vehicle_journey in vehicle_journeys_for_line
]
```

repos/fdbt-reference-data-service/src/uploaders/txc_uploader/txc_processor.py
```diff
@@ -177,45 +186,107 @@ def collect_journey_patterns(data: dict, service: dict):
...
     for journey_pattern in journey_patterns:
+        journey_pattern_info = journey_pattern["journey_pattern_info"]
+
+        if (
+            journey_pattern_info["journey_pattern_ref"]
+            not in vehicle_journey_journey_pattern_refs
+        ):
+            continue
+
+        if check_journey_pattern_exists(
```
ljones92, line 249: I'd maybe remove this check and the `check_journey_pattern_exists` function, it might lead to some issues since it's not including the section refs like disruptions does

repos/fdbt-reference-data-service/src/uploaders/txc_uploader/txc_processor.py
alainajaffer: done

repos/fdbt-reference-data-service/src/uploaders/txc_uploader/txc_processor.py
alainajaffer: done

ljones92 (approved):

PR #936, opened 2024-12-05
fix: only map over valid products in exporter

nbetts:
## Description

The select exports page displays a count of all the user's products available for export, however this count includes invalid products that are not available to select. The fix is to omit invalid products from the count.

## Testing instructions

In pre-prod, the FSYO operator contains two invalid products and several valid products. Assert that the count of selected products matches the number of available products to select in the export tool. The count is in an indicator tag in the top-left of the page, e.g. "0 / 5 selected"

### DIFF
--- a/repos/fdbt-site/src/pages/products/selectExports.tsx
+++ b/repos/fdbt-site/src/pages/products/selectExports.tsx
@@ -482,11 +482,15 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
     const allServicesWithMatchingLineIds: Record<string, MyFaresService> = {};
 
     const validProducts = nonExpiredProducts.filter((product) => {
+        if (!product.lineId) {
+            return true;
+        }
+
+        // If the product has a line ID but the ID doesn't match any of the services, the product should no longer be surfaced
         const matchingService = allBodsServices.find((service) => service.lineId === product.lineId);
 
         if (matchingService) {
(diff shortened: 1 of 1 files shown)


repos/fdbt-site/src/pages/products/selectExports.tsx
```diff
@@ -474,12 +474,28 @@ export const getServerSideProps = async (ctx: NextPageContextWithSession): Promi
...
     const nonExpiredProducts = getNonExpiredProducts(products);
-
     const allPassengerTypes = await getAllPassengerTypesByNoc(noc);
 
+    const multiModalAttribute = getSessionAttribute(ctx.req, MULTI_MODAL_ATTRIBUTE);
+    const dataSource = multiModalAttribute ? 'tnds' : 'bods';
+    const allBodsServices = await getBodsOrTndsServicesByNoc(noc, dataSource);
+    const allServicesWithMatchingLineIds: Record<string, MyFaresService> = {};
+
+    const validProducts = nonExpiredProducts.filter((product) => {
```
DeannaSharma, line 484:
```suggestion
    const validProducts = nonExpiredProducts.filter((product) => {
    
      if (!product.lineId) {
            return true;
        }
```

I noticed some behaviour where any fare that wasn't a service-type one was getting filtered out and couldn't be selected for exporting.

I think to resolve this we need a check so that if lineId doesn't exist in the product table we count it as a valid product to display.

DeannaSharma (approved):

PR #947, opened 2025-02-04
CFDS-317 multi-operator external journey (lead operator only)

nbetts:
## Description

Amends the existing create product journey to allow creating multi-operator external products

## Testing instructions

- Test creating a multi-operator external fare zone product and assert the products gets created with status incomplete (since any additional operators won't have filled their part of the product)
- Test creating a multi-operator external services product and assert the products gets created with status incomplete (since any additional operators won't have filled their part of the product)
- Regression test the other product type journeys and assert they still work correctly

### DIFF
--- a/repos/fdbt-site/src/pages/ticketConfirmation.tsx
+++ b/repos/fdbt-site/src/pages/ticketConfirmation.tsx
@@ -29,6 +29,7 @@ import {
 } from '../constants/attributes';
 import {
     ConfirmationElement,
+    FareType,
     MultipleOperatorsAttribute,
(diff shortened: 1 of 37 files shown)


repos/fdbt-site/src/pages/ticketConfirmation.tsx
DeannaSharma, line 1: On this page, I don't believe we want the "change" link to appear for the secondary ops. Clicking the link takes me to a blank page
Screenshot 2025-02-05 at 09 02 08

DeannaSharma (approved):

PR #949, opened 2025-02-06
CFDS-316-product-details-page-changes

nbetts:
## Description

- amend product details page to display for multi-operator prodect non-lead operators (services variation only)


repos/fdbt-site/src/pages/products/productDetails.tsx
DeannaSharma, line 1: On line 120, we also need to amend the edit link next to the product name for non lead ops
image

DeannaSharma (approved):

PR #961, opened 2025-02-14
feat: Export page improvements

DeannaSharma:
## Description

Improvements to multiop dashboard and export page to show a notification banner when an export is in progress. Buttons for exporting will no longer be hidden if an export is in progress, instead an error will be displayed if the button is clicked when an export is in progress.


repos/fdbt-site/src/pages/products/exports.tsx
```diff
@@ -81,47 +83,87 @@ const Exports = ({ csrf, operatorHasProducts, initialExportStarted }: GlobalSett
...
+            }
+        }
+    };
+
     return (
         <>
             <BaseLayout title={title} description={description} showNavigation>
                 <div className="govuk-grid-row">
+                    {!!exportErrorState && (
+                        <>
```
nbetts, line 114: fragments are only needed when rendering multiple top-level elements

nbetts (approved):

PR #974, opened 2025-02-28
feat: new lambda for sending multi-operator email notifications

nbetts:


DeannaSharma (dismissed):

repos/fdbt-emailer/src/multi-operator-emailer/email.ts
```diff
@@ -0,0 +1,89 @@
...
+export const sendEmails = async (
+    client: SES,
+    serviceDomain: string,
+    serviceEmailAddress: string,
+    emailAddresses: string[],
+): Promise<void> => {
+    const mailTransporter = nodemailer.createTransport({ SES: client });
+
+    const mailOptions: Options[] = emailAddresses.map((emailAddress) => ({
+        from: serviceEmailAddress,
```
DeannaSharma, line 80:
```suggestion
        from: `no-reply@${serviceDomain}`,
```

With the changes to SES I think this can be updated as above. This might not work for prod until the gov.uk domain is verified in SES

repos/fdbt-emailer/src/multi-operator-emailer/handler.ts
```diff
@@ -0,0 +1,28 @@
+import 'source-map-support/register';
+import { Handler } from 'aws-lambda';
+import { getCognitoClient, getUsersOptedIntoMultiOperatorEmails } from './cognito';
+import { getAuroraDBClient, getIncompleteMultiOperatorProducts } from './database';
+import { getSesClient, sendEmails } from './email';
+
+export const handler: Handler = async () => {
```
DeannaSharma, line 7: I think there's a couple situations we could try to cover:
- If there are no users opted into multiop emails we should end the function early without making a call to the DB to get products
- If `emailAddresses` is an empty array it might be best to end the function early and not attempt to send any emails

nbetts: I'm going to switch it and do the DB check first as it is potentially fewer network calls

DeannaSharma (approved):

PR #1014, opened 2026-04-30
DBODS-1113: Vulnerability bumps

HarrisonDuffield:
## Description

- Requests Lib:

Increases value of requests library from 2.32.2 to 2.33.0. 
https://github.com/psf/requests/blob/main/HISTORY.md

There are no breaking changes indicated in release notes above

- xmltodict
https://github.com/martinblech/xmltodict/blob/master/CHANGELOG.md
No breaking changes in versions. (Minor version bump only as this is brought in to the ticket as scope creep)

- pytest

Testing lib only - no breaking changes that will effect 

- axios (transitive)
- fast-xml-parser (transitive)

## Testing instructions

This should be tested via the ticket previous that this will unblock and via regression

### DIFF
--- a/repos/fdbt-netex-output/package-lock.json
+++ b/repos/fdbt-netex-output/package-lock.json
@@ -84,613 +84,758 @@
                 "node": ">=6.0.0"
             }
         },
-        "node_modules/@aws-crypto/crc32": {
-            "version": "3.0.0",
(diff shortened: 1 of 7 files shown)


cpaul92 (approved):
Changes look good. Do we need to have a conversation about how this will be tested?

PR #1019, opened 2026-05-18
revert(ci): undo safety version pin in reference-data-service workflows

m-a-r-i-a-n:
## Description

Explain the purpose of this PR

## Testing instructions

Add testing instructions, if required


prolude-dks (approved):
Non-blocking: the same Safety setup is repeated across multiple workflows. Longer term, it may be worth moving this into a reusable workflow or composite action to reduce drift.

Subject to the checks passing, I am happy with the intent of the change.

PR #1038, opened 2026-06-29
fix: csv uploader lambda recursive loop

sophieoneill:
## Description

This PR is to fix the code causing a recursive loop in the `reference-data-service-uploaders-test-CsvUploader` lambda. The lambda was being caused due to the added NaPTAN handler adding a file to the staging S3 bucket, which is the trigger to the overall lambda.

I have moved the NaPTAN functionality that puts the file in a staging S3 bucket from the uploader lambda to the retriever lambda, which is called on a daily schedule, removing that recursive loop.

https://busopendataservice.atlassian.net/browse/DBODS-2001

### DIFF
--- a/repos/fdbt-reference-data-service/src/uploaders/csv_uploader/db_queries.py
+++ b/repos/fdbt-reference-data-service/src/uploaders/csv_uploader/db_queries.py
@@ -28,11 +28,11 @@ def noc_lines_query(bucket_name):
     ]
 
 
-def stops_query(bucket_name):
+def stops_query(bucket_name, naptan_s3_key, naptan_bucket_region):
     return [
         "SET FOREIGN_KEY_CHECKS = 0;",
         "DROP TABLE IF EXISTS naptanStopNew;",
         "CREATE TABLE naptanStopNew LIKE naptanStop;",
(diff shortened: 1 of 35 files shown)


repos/fdbt-reference-data-service/src/uploaders/csv_uploader/db_queries.py
```diff
@@ -28,11 +28,11 @@ def noc_lines_query(bucket_name):
...
+        f"LOAD DATA FROM S3 's3-eu-west-2://{bucket_name}/Stops.csv' REPLACE INTO TABLE naptanStopNew CHARACTER SET UTF8 FIELDS TERMINATED BY ',' OPTIONALLY ENCLOSED BY '\"' LINES TERMINATED BY '\\n' IGNORE 1 lines ( atcoCode, naptanCode, plateCode, cleardownCode, commonName, commonNameLang, shortCommonName, shortCommonNameLang, landmark, landmarkLang, street, streetLang, crossing, crossingLang, indicator, indicatorLang, bearing, nptgLocalityCode, localityName, parentLocalityName, grandParentLocalityName, town, townLang, suburb, suburbLang, localityCentre, gridType, easting, northing, longitude, latitude, stopType, busStopType, timingStatus, defaultWaitTime, notes, notesLang, administrativeAreaCode, creationDateTime, modificationDateTime, revisionNumber, modification, status );",
```
HarrisonDuffield, line 36: Just clarifying are we committing to no env vars for these fields?

I know we won't ever leave euwest2 but just making sure

HarrisonDuffield (approved):

repos/fdbt-reference-data-service/src/uploaders/csv_uploader/db_queries.py
a-joglekar, line 36: Further to Harrison's comment, do env variables added for retriever lambda also available in uploader lambda? If so, we should use them to parameterise and ensure consistency

sophieoneill: Env vars will be added to the retriever same as they had been in uploader. I think bucket region will never change, and I have changed it to eu-west-2 here as that is what is used on all other queries in this script, so keeping it consistant with that.

sophieoneill: I changed the bucket name to be hard coded as like you said we won't be changing that, plus that is how all the other queries in this script call the data so wanting to keep it consistant. Also it's not calling the naptan bucket anymore, so moving away from the naptan env vars in the uploader.

PR #1041, opened 2026-06-30
Fix: DBODS-1629: Updating to read NOC data from S3 bucket

CaelainnBrinkley:
Updating NOC Code reference to call S3 bucket from BODS instead of calling directly from the API
Updates to the following files:

- repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
- repos/fdbt-reference-data-service/src/retrievers/serverless.yml

This PR is following the revert of the last create data PR https://github.com/department-for-transport-BODS/bods-transport-data-parser/pull/359 which was reverted due to the catch of a recursive loop that was implemented by mistake

### DIFF
--- a/repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
+++ b/repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
@@ -145,7 +145,7 @@ def noc_handler(event, context):
             noc_s3_keys = stage_noc_file_locally(noc_bucket, noc_role_arn, noc_bucket_region, local_bucket)
             bucket = local_bucket
         else:
-            # List objects in the folder and select the 3 required CSVs directly
+            # List objects in the folder and select the 3 required CSVs directly.
             s3_client = boto3.client('s3', region_name=noc_bucket_region)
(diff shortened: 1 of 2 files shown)


repos/fdbt-reference-data-service/src/retrievers/csv_retriever/main.py
```diff
@@ -69,6 +69,116 @@ def naptan_handler(event, context):
...
+                s3_client.copy_object(
+                    Bucket=noc_bucket,
+                    CopySource={"Bucket": noc_bucket, "Key": source_key},
+                    Key=target_filename,
+                )
+                noc_s3_keys.append(target_filename)
+            bucket = noc_bucket
+
+        # Process each selected CSV
+        for noc_s3_key in noc_s3_keys:
```
a-joglekar: The processing of the files should be done via csv_uploader lambda. Within retriever, we should only stage the files locally. So please move this call into the retriever lambda. Since the files are locally staged now, the existing function can be used, similar to way Sophie has done for NaPTAN

a-joglekar (changes requested):

a-joglekar (approved):

PR #1053, opened 2026-09-04
Add preprod Cypress test suite with HTML reporting

naimullakainos:
Copies the existing e2e specs into `cypress/e2e/preprod` and runs them against
https://preprod.dft-cfd.com using a dedicated `cypress.preprod.config.ts`.

Preprod does not allow the `disableAuth` bypass, so the tests sign in through the
login page. Two accounts are used: a scheme operator for `scheme.cy.ts` and
`carnet/scheme.cy.ts`, and a normal operator for everything else. Credentials come
from environment variables and are not committed.

- Preprod-only fare zone fixtures using ATCO codes that exist on the environment
- Test data namespaced with a `Preprod Cypress` prefix
- Scheme specs run their own setup, including operator details
- Mochawesome HTML report via `npm run runCypress:preprod:report`
- Fixes for two genuine test bugs: the "Select All Services" toggle could clear the
  selection when editing, and fare stage matching assumed a fixed number of stops
- README updated

All preprod-specific behaviour is guarded by `Cypress.env('preprod')`, so local runs
are unchanged.

Latest run: 58 tests, 53 passed, 5 pending, 0 failures.

Not run on preprod: flat fare priced by distance (needs `STAGE=dev`) and the four
global settings lifecycle specs (they delete data that products depend on).

Test defects found and fixed
These were genuine bugs in the shared test code, and the fixes benefit local runs too:

1. Service selection could end up empty. "Select All Services" is a toggle whose label is initialised from the product's saved state. When editing a product that already had all services, clicking it cleared the selection, so the form was submitted with none selected. Now verified and corrected after selection.
2. Fare-stage matching assumed a fixed number of stops. It hard-coded four dropdowns and index 2. Preprod services have different stop counts. Matching now assigns every fare stage and uses the page's auto-populate control.
3. Transient error page aborted whole specs. A one-off 500 on /viewOperatorGroups killed a spec because Cypress never retries before all hooks. A reload guard was added.


repos/fdbt-site/cypress_tests/package.json
```diff
@@ -5,6 +5,10 @@
     "scripts": {
         "openCypress": "cypress open --config watchForFileChanges=false",
         "runCypress": "cypress run -b $BROWSER",
+        "runCypress:preprod": "cypress run --config-file cypress.preprod.config.ts -b $BROWSER",
+        "runCypress:preprod:report": "rm -rf results/mochawesome results/preprod-report.html results/preprod-report.json && cypress run --config-file cypress.preprod.config.ts -b $BROWSER --reporter mochawesome --reporter-options reportDir=results/mochawesome,overwrite=false,html=false,json=true; npm run report:build && npm run report:open",
+        "report:build": "mochawesome-merge results/mochawesome/*.json > results/preprod-report.json && marge results/preprod-report.json -f preprod-report -o results --inline",
+        "report:open": "open -a 'Google Chrome' results/preprod-report.html",
```
HarrisonDuffield: Open -a will break on windows.

I think it will also break on linux - or anyone without chrome.

---
I'd be tempted just to print the file path in the terminal and the user can open it themselves.

repos/fdbt-site/cypress_tests/cypress/e2e/preprod/globalSettings/operatorDetails.cy.ts
```diff
@@ -0,0 +1,12 @@
+import { startGlobalSettings } from '../../../support/globalSettings';
+import { enterOperatorDetails } from '../../../support/operatorDetails';
+
+describe('time restrictions', () => {
+    it('creates edits and deletes time restrictions', () => {
```
HarrisonDuffield: This test doesn't set any time restrictions

repos/fdbt-site/cypress_tests/cypress/e2e/preprod/carnet/flatFare.cy.ts
```diff
@@ -0,0 +1,28 @@
...
+        defineUserTypeAndTimeRestrictions();
+        completeFlatFareCarnet();
+        completeSalesPages(3, 'Flat fare carnet ');
+        isFinished();
+    });
+
+    it('completes successfully for geoZone', () => {
+        selectCarnetFareType('flatFare');
+        defineUserTypeAndTimeRestrictions();
+        clickElementById('radio-option-geoZone');
```
HarrisonDuffield: is it worth making selecting these a step/component rather than referring to it by ID. It's here a few times

repos/fdbt-site/cypress_tests/cypress/e2e/preprod/myFares/myFares.cy.ts
```diff
@@ -0,0 +1,261 @@
...
+    it("allows the user to edit a product's name", () => {
+        getHomePage();
+        clickElementById('account-link');
+        clickElementByText('Other products');
+        clickRandomElementInTable('govuk-table__body', 'product-link');
+        editProductName();
+    });
+    it("allows the user to edit a product's validity", () => {
+        getHomePage();
+        clickElementById('account-link');
```
HarrisonDuffield: Again this comes up a bunch, would it be more maintainable going forwars to have the account-link as a component that can be updated in one place.

Same anywhere else this pattern happens

repos/fdbt-site/cypress_tests/cypress/support/helpers.ts
```diff
@@ -50,7 +50,43 @@ export const clearAndTypeByName = (name: string, text: string): void => {
 
 export const getRandomNumber = (min: number, max: number): number => Cypress._.random(min, max);
 
+export const getTestDataName = (name: string): string => (Cypress.env('preprod') ? `Preprod Cypress ${name}` : name);
+
+// Deployed environments occasionally return the generic error page on a page load, and a
+// failure inside a `before all` hook is never retried, so reload before reading the page.
+export const reloadOnServiceError = (): void => {
```
HarrisonDuffield, line 62: Can we raise a ticket to look into this pleae

HarrisonDuffield (changes requested):
minor changes

repos/fdbt-site/cypress_tests/package.json
naimullakainos: Addressed in commit 41d17634b. The report command now prints the generated HTML report path instead of using macOS-only `open -a`, and the README has been updated accordingly.

repos/fdbt-site/cypress_tests/cypress/e2e/preprod/globalSettings/operatorDetails.cy.ts
naimullakainos: Addressed in commit 6cde70eea. Corrected the suite and test names to `operator details` and `updates operator details`, matching what the test actually exercises.

repos/fdbt-site/cypress_tests/cypress/e2e/preprod/carnet/flatFare.cy.ts
naimullakainos: Addressed in commit 6cde70eea. Added the shared `selectProductScope()` helper and used it for the repeated product-scope selectors.

repos/fdbt-site/cypress_tests/cypress/e2e/preprod/myFares/myFares.cy.ts
naimullakainos: Addressed in commits 6cde70eea and 41d17634b. Added `openAccountSettings()` and replaced repeated `account-link` selectors across the local/preprod my-fares specs and shared setup paths.

repos/fdbt-site/cypress_tests/cypress/support/helpers.ts
naimullakainos: The reload guard remains in this PR, but I have not created a separate ticket yet. This is a follow-up investigation for transient service errors rather than a code change covered by the pushed commits.

naimullakainos: Raised a ticket for this https://busopendataservice.atlassian.net/browse/DBODS-3031

naimullakainos:
Updated Cypress pre-production global-settings tests to support unique, parameterized test data and avoid conflicts between test runs.

Changes

Added reusable global-settings navigation helpers.
Improved element selection using card content instead of fixed card indexes.
Added configurable name prefixes for:
Passenger types
Passenger groups
Purchase methods
Time restrictions
Multi-operator groups
Updated helper functions to support dynamic passenger names and group data.
Updated all related pre-production Cypress specs.

Tests passed except the flat fare priced by distance (needs STAGE=dev) which was agreed to skip
image

HarrisonDuffield (approved):

